PR #3, opened 2022-10-05
Fix default boolean juju config

gregory-schiano:
Since ops 1.5.3, the config check is stricter see
https://github.com/canonical/operator/pull/787

This PR intends to fix an issue in the enable_cors default value


arturo-seijas (approved):
LGTM

mthaddon (approved):
LGTM, thx

jdkandersson:
Thanks for fixing this, I have noticed that, when interacting with the configuration in charm code, you have to pass a string value for boolean fields I believe, otherwise juju throws an error

PR #4, opened 2022-10-18
Feature/isd 226 add integration tests

gregory-schiano:
Added 4 integration tests:
- Check charm is deployed
- Check service is working
- Register and admin, activate the account and get an API Key
- Change config to enable S3 and check buckets is created and contains files

In order to implement integration tests I needed to do some refactoring on the charm itself:
Previously the charm was running the "setup" process in the start script, leading to the charm being declared as available but the service not yet running.
In order to avoid this situation, the setup script is now executed during the "pebble" ready event, and services is effectively started when the charm is active

The S3 test assumes a [localstack](https://localstack.cloud/) is available. It supports localstack running locally or remotely
To run localstack execute the following command:
`docker run -d --restart=always -ti --name localstack -p 80:4566 localstack/localstack:latest`
Beware of the port "80" it is required as discourse needs an S3 service running on port 80

To run the integration tests you'll first need to rebuild the discourse docker images due to the changes in the startup scripts:
```
make build-image
make push-image-local-registry
```

Then run the following command:
`$ENV_FOLDER/bin/tox -r -e integration -- --model $MODEL_NAME --discourse-image localhost:32000/discourse:latest`
Where:
- $ENV_FOLDER is the name of your virtual env
- $MODEL_NAME is the name of the juju model you will use to deploy charms for the tests

If localstack runs on a remote machine please use:
`--s3-ip-address $IP_ADDRESS`
Where $IP_ADDRESS is the IP address of the host running localstack


.gitignore
```diff
@@ -4,3 +4,8 @@
 .coverage
 __pycache__
 build
+.vscode
+.env
+venv
+env
+.pytest_cache
```
arturo-seijas, line 11: One picky comment here. There's a blank line missing at the end

.jujuignore
```diff
@@ -1,6 +1,9 @@
-/env
+/venv
 /image
 *.py[cod]
 *.charm
 *.tar.gz
 Makefile
+/.vscode
+.env
+/.pytest_cache
```
arturo-seijas, line 9: Same here

src/charm.py
```diff
@@ -202,7 +213,8 @@ def _get_s3_env(self):
             "DISCOURSE_S3_ENDPOINT": self.config.get("s3_endpoint", "s3.amazonaws.com"),
             "DISCOURSE_S3_REGION": self.config["s3_region"],
             "DISCOURSE_S3_SECRET_ACCESS_KEY": self.config["s3_secret_access_key"],
-            "DISCOURSE_USE_S3": True,
+            "DISCOURSE_S3_INSTALL_CORS_RULE": str(self.config.get("s3_install_cors_rule", True)).lower(),
+            "DISCOURSE_USE_S3": "true",
```
arturo-seijas, line 217: Why this change?

src/charm.py
```diff
@@ -202,7 +213,8 @@ def _get_s3_env(self):
             "DISCOURSE_S3_ENDPOINT": self.config.get("s3_endpoint", "s3.amazonaws.com"),
             "DISCOURSE_S3_REGION": self.config["s3_region"],
             "DISCOURSE_S3_SECRET_ACCESS_KEY": self.config["s3_secret_access_key"],
-            "DISCOURSE_USE_S3": True,
+            "DISCOURSE_S3_INSTALL_CORS_RULE": str(self.config.get("s3_install_cors_rule", True)).lower(),
```
arturo-seijas, line 216: Can't you use a boolean instead?

src/charm.py
```diff
@@ -274,14 +292,33 @@ def _create_layer_config(self):
...
+                    "override": "replace",
+                    "http": {"url": f"http://localhost:{SERVICE_PORT}"},
+                },
+            }
         }
         return layer_config
 
+    def _should_run_setup(self):
+        # Return true if setup never ran
+        return not self._stored.setup_ran or (
```
arturo-seijas, line 311: Usage of stored state is discouraged https://juju.is/docs/sdk/stored-state-uses-limitations. Is it possible to retrieve the previous value from another source?

src/charm.py
```diff
@@ -302,6 +339,33 @@ def _config_changed(self, event):
...
+                stdout, _ = process.wait_output()
+                logger.debug(f"{script} stdout: {stdout}")
+            except ExecError as e:
+                logger.error(f"{script} command exited with code {e.exit_code}. Stderr:")
+                for line in e.stderr.splitlines():
+                    logger.error(f"    {line}")
+                logger.debug(f"{script} stdout: {e.stdout}")
+                self.model.unit.status = BlockedStatus(f"Error while executing {script}")
+                raise
+            self._stored.setup_ran = True
```
arturo-seijas, line 359: Same here

arturo-seijas (changes requested):

src/charm.py
gregory-schiano: Env vars are only strings, and at some point I got issue due to that, so decided to force string for all env vars

src/charm.py
gregory-schiano: I discussed this topic with Tom, indeed not the best way.
Unfortunately I don't see any other possibility to know if this script has previously run or not.
Note that the script is "idempotent" so could basically be run for every event linked to this callback, but not sure that's what we want

.gitignore
gregory-schiano: Done

.jujuignore
gregory-schiano: Done

src/charm.py
arturo-seijas: Understood

src/charm.py
arturo-seijas: Considering this will be ran only when the config changed event, I'd personally go for running it every time the callback is triggered. Configuration changes are not very frequent so I don't think performance will be an issue here.

weiiwang01 (dismissed):

src/charm.py
gregory-schiano: The name of the method is misleading, it's bound to multiple events:
```
        self.framework.observe(self.on.leader_elected, self._config_changed)
        self.framework.observe(self.on.upgrade_charm, self._config_changed)
        self.framework.observe(self.on.redis_relation_updated, self._config_changed)
```
Don't know frequency of occurence for these events.

What I could eventually do is write a file on the leader filesystem to store the "state"
But it won't be persisted when the containers is recreated

What do you think ?

tests/integration/test_charm.py
```diff
@@ -0,0 +1,283 @@
...
+    logger.info("Updating discourse hosts")
+    
+    # Discourse S3 client uses subdomain bucket routing,
+    # I need to inject subdomain in the DNS (not needed if everything runs localhost)
+    action = await app.units[0].run(f"echo \"{s3_ip_address}  {s3_bucket}.{s3_domain}\" >> /etc/hosts")
+    result = await action.wait()
+    assert result.results.get("return-code") == 0, "Can't inject S3 IP in Discourse hosts"
+
+    logger.info("Injected bucket subdomain in hosts, configuring settings for discourse")
+
```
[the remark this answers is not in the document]
gregory-schiano: The localstack must be reset after every run, in a next step I'll add github action which will deploy a localstack for every integration run
So I don't think (unless for testing purpose), we need to cleanup the bucket in the integration test itself

tests/integration/test_charm.py
```diff
@@ -0,0 +1,283 @@
...
+    
+    # Check the bucket has been created
+    response = s3_client.list_buckets()
+    bucket_list = [*map(lambda a: a["Name"], response["Buckets"])]
+    
+    assert s3_bucket in bucket_list
+    
+    # Check content has been uploaded in the bucket
+    response = s3_client.list_objects(Bucket=s3_bucket)
+    bucket_count = sum(1 for _ in response["Contents"])
```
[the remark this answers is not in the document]
gregory-schiano: Done

weiiwang01 (dismissed):

src/charm.py
arturo-seijas: It seems to me that we can achieve a simpler working solution with some additional refactoring. So I say we park this for now.

From what I see in https://github.com/canonical/discourse-k8s-operator/blob/main/image/scripts/pod_setup, redis is only used to acquire a lock, hence, there's no need to run the hook on `redis_relation_updated`.
Both `discourse_pebble_ready`and `upgrade_charm`won't be triggered without a `config_changed` event, so there's no need for those callbacks, also, I don't see why running it again when leadership changes.

We can get the env var values that really require the execution of the script (probably all used in it) from the containers plan by running `Container.get_plan` and comparing it to the charm config.

arturo-seijas: I had the same thought, however, leadership is only guaranteed for 30s and it can be changed during the hook execution, so I'm not sure removing it is safe.

gregory-schiano: So our discussion gave me an idea, I refactored to use the presence of a pebble plan to launch the startup script, and next to it I use the pebble plan env and compare it with the charm config to know if any value has changed

arturo-seijas (dismissed):
LGTM

weiiwang01 (dismissed):

weiiwang01 (approved):

arturo-seijas (approved):
LGTM

PR #5, opened 2022-10-20
Fix unit tests failing due to bad usage of "add_relation"

gregory-schiano:
Due to a bad usage of "add_relation" method of Harness, the unit tests started to fail with ops 1.5.3 (which is more strict in validating data relations, see canonical/operator#786)


arturo-seijas (approved):
LGTM

gtrkiller (approved):
LGTM

PR #6, opened 2022-11-10
Add workflows and issue templates

arturo-seijas:


merkata (dismissed):
Nice! The lint and coverage are failing, but it is thanks to the workflows that it is apparent, so I'm approving the addition (and the failing tests can be fixed on subsequent PRs).

gtrkiller (dismissed):
Going to approve, but note that the integration tests are failing, so please fix them in a future PR please.

pyproject.toml
```diff
@@ -1,12 +1,17 @@
...
+[tool.bandit.assert_used]
+skips = ["*/*test.py", "*/test_*.py"]
+
 # Testing tools configuration
 [tool.coverage.run]
 branch = true
 
 [tool.coverage.report]
-fail_under = 86
+fail_under = 83
```
mthaddon, line 14: We have a team policy of trying to stay above 85% test coverage. Can we have a look at preserving that here?

src/charm.py
```diff
@@ -344,17 +356,25 @@ def _config_changed(self, event):
...
         # - Change in important S3 parameter (comparing value with envVars in pebble plan)
-        if self._check_config_is_valid() and self.model.unit.is_leader() and self._should_run_setup(current_plan, previous_s3_info):
+        if (
+            self._check_config_is_valid()
+            and self.model.unit.is_leader()
+            and self._should_run_setup(current_plan, previous_s3_info)
+        ):
             script = f"{SCRIPT_PATH}/pod_setup"
 
             logger.debug(f"Executing setup script ({script})")
```
mthaddon, line 374: We shouldn't use f-strings for logging.

.flake8
```diff
@@ -0,0 +1,14 @@
...
+  venv
+  .git
+  build
+  dist
+  lib/charms
+  *.egg_info
+ignore:
+  E501
+  W503
+
```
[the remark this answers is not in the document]
arturo-seijas: Indeed

pyproject.toml
arturo-seijas: I wasn't planning to add more tests as part of this PR but I cna leave it at 85 so that the wf will fail

arturo-seijas: I'll add some tests

gtrkiller (commented):
Looks good overall, but you may want to address the shellcheck test failure, as it is throwing several warnings on the actions.

weiiwang01 (dismissed):

.flake8
```diff
@@ -0,0 +1,13 @@
...
+select: E,W,F,C,N
+exclude:
+  venv
+  .git
+  build
+  dist
+  lib/charms
+  *.egg_info
+ignore:
+  E501
```
jdkandersson, line 12: What is the reason for these ignores please?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,57 @@
+name: Bug Report
```
jdkandersson, line 1: Similar to the operator workflows, can this be shared across or repositories?

image/build_scripts/get_app_dependencies
```diff
@@ -34,4 +34,4 @@ apt-get install -y brotli \
 # base image is >= 22.04. Also, please consider removing the node-uglify
 # when upgrading to discourse >= 2.8.0, since the node-uglify is not
 # required to trigger node js assets compression, only terser will do fine.
-which terser || ln -s $(which uglifyjs.terser) /usr/local/bin/terser
\ No newline at end of file
+which terser || ln -s "$(which uglifyjs.terser)" /usr/local/bin/terser
```
jdkandersson, line 37: Should there be a new line here?

pyproject.toml
```diff
@@ -38,4 +44,4 @@ docstring-convention = "google"
 # Check for properly formatted copyright header in each file
 copyright-check = "True"
 copyright-author = "Canonical Ltd."
-copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
\ No newline at end of file
+copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
```
jdkandersson, line 47: Do you think there is a way to move this regex to the operator workflows? Maybe something like we did for the woke tool?

tox.ini
```diff
@@ -22,7 +22,40 @@ passenv =
...
+commands =
+    # uncomment the following line if this charm owns a lib
+    # codespell {[vars]lib_path}
+    codespell {toxinidir}/. --skip {toxinidir}/.git --skip {toxinidir}/.tox \
+      --skip {toxinidir}/build --skip {toxinidir}/lib --skip {toxinidir}/venv \
+      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg
+    # pflake8 wrapper supports config from pyproject.toml
+    pflake8 {[vars]all_path}
+    isort --check-only --diff {[vars]all_path}
+    black --check --diff {[vars]all_path}
```
jdkandersson, line 55: Should we also add pylint and mypy?

.github/ISSUE_TEMPLATE/bug_report.yml
arturo-seijas: It can be added automatically if defined in the .github repository, but unfortunately there's no way to add it only to a subset of projects. This is charm specific, so we'll need to find an alternative approach.

tox.ini
arturo-seijas: The tox configuration is shared accross charms, so it'd better be done here https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-simple/tox.ini.j2

image/build_scripts/get_app_dependencies
arturo-seijas: Indeed. Thx

.flake8
arturo-seijas: E501 is "line too long", but splitting the affected lines is causing `black` to fail
```
-            "https://login.ubuntu.com/+saml":
-                "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"
+            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"
```
```
"command":
-                        f"sh -c '{SCRIPT_PATH}/pod_start >>/srv/discourse/discourse.log 2&>1'",
+                    "command": f"sh -c '{SCRIPT_PATH}/pod_start >>/srv/discourse/discourse.log 2&>1'",

```
W503 is "line break before binary operator". This has been ignored as it is the recommended way to split lines as in 
```
self.config.get("s3_enabled")
            and (
                s3info.enabled != self.config.get("s3_enabled")
                or s3info.region != self.config.get("s3_region")
                or s3info.bucket != self.config.get("s3_bucket")
                or s3info.endpoint != self.config.get("s3_endpoint")
            )
```

pyproject.toml
arturo-seijas: Yes, good point. Maybe we can file a ticket for this, what do you think?

merkata (dismissed):
Looks great, good additions. Approving.

merkata (dismissed):

.flake8
jdkandersson: Ah, do you think it is worth adding an explanatory comment above the ignores so that others know these reasons in the future?

tox.ini
jdkandersson: We can leave this for another PR, we have already deviated from the `tox.ini` file by adding black for a good reason that we needed to have additional security checks. A similar argument could apply to `pylint`, `mypy` and potentially `pydocstyle` if we believe there is value in the additional checks

pyproject.toml
jdkandersson: Created a ticket

tox.ini
arturo-seijas: Indeed, but we should probably look into converging again by contributing there

tests/unit/test_charm.py
```diff
@@ -160,11 +160,66 @@ def test_config_changed_when_s3_and_no_bucket_invalid(self):
...
+        self.assertEqual(
+            "discourse.local", self.harness.charm.ingress.config_dict["service-hostname"]
+        )
+
     def test_config_changed_when_valid(self):
         """
         arrange: given a deployed discourse charm with all the required relations
         act: when a valid configuration is provided
-        assert: the approapriate configuration values are passed to the pod and the unit reaches Active status.
+        assert: the approapriate configuration values are passed to the pod and the unit
```
mthaddon, line 221: s/approapriate/appropriate/

mthaddon (dismissed):
One very minor comment but overall this looks good to me.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,57 @@
...
+    attributes:
+      value: >
+        Thanks for taking the time to fill out this bug report! Before submitting your issue, please make
+        sure you are using the latest version of the charm. If not, please switch to this image prior to 
+        posting your report to make sure it's not already solved.
+  - type: textarea
+    id: bug-description
+    attributes:
+      label: Bug Description
+      description: >
```
gtrkiller, line 15: Why are we using > instead of putting the description in a single line?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,57 @@
...
+        If applicable, add screenshots to help explain the problem you are facing.
+    validations:
+      required: true
+  - type: textarea
+    id: reproduction
+    attributes:
+      label: To Reproduce
+      description: >
```
gtrkiller, line 23: Same as above

gtrkiller (changes requested):
Some comments inline, requesting changes because of the test coverage decrease mostly.

pyproject.toml
```diff
@@ -1,12 +1,17 @@
...
+[tool.bandit.assert_used]
+skips = ["*/*test.py", "*/test_*.py"]
+
 # Testing tools configuration
 [tool.coverage.run]
 branch = true
 
 [tool.coverage.report]
-fail_under = 86
+fail_under = 85
```
gtrkiller, line 14: Why are we decreasing the minimum coverage?

arturo-seijas: When changing the conf files the reported coverage decreased. I've added some tests to reach the minimum per our standards.

.github/ISSUE_TEMPLATE/bug_report.yml
arturo-seijas: To have multiline strings, so that they'll fit in a iff

.github/ISSUE_TEMPLATE/bug_report.yml
arturo-seijas: To have multiline strings, so that they'll fit in a iff

gtrkiller (dismissed):
LGTM

.flake8
arturo-seijas: I think the explanation and it'll be difficult to provide more context than what the rule is about

.woke.yaml
```diff
@@ -0,0 +1,12 @@
+rules:
+  # Ignore "whitelist" - upstream uses this terminology and it would be
+  # confusing to change in this charm while upstream continues to use it.
+  # Specifically the following annotations:
+  #   - nginx.ingress.kubernetes.io/limit-whitelist
+  #   - nginx.ingress.kubernetes.io/whitelist-source-range
+  # See https://github.com/kubernetes/ingress-nginx/issues/7916 for upstream
+  # issue about this.
+  - name: whitelist
```
mthaddon, line 9: Can we rather just ignore the library file? I think it would be good to do that rather than ignore whitelist anywhere else in the code.

.woke.yaml
```diff
@@ -0,0 +1,12 @@
...
+  # Ignore "master" - the databse relation event received from the library
```
mthaddon, line 10: s/databse/database/

arturo-seijas: Fixed

tox.ini
jdkandersson: Let's continue this discussion outside the PR

gregory-schiano (dismissed):
It's been a fight to have integration test successful but all good to me !

tests/integration/test_charm.py
```diff
@@ -85,7 +85,7 @@ async def test_setup_discourse(
     # Parse output and send registration form
     parsed_registration: BeautifulSoup = BeautifulSoup(response.content, features="html.parser")
 
-    # Get the form info
+`    # Get the form info
```
arturo-seijas: Is this a typo?

gregory-schiano: That was, I removed it normally

mthaddon (approved):
Hefty changes here! Thanks for pushing through with this, looks good to me.

weiiwang01 (approved):

PR #9, opened 2022-11-16
Refactor dockerfile

arturo-seijas:
Some refactoring for better layering of the dockerfile and increased readability. Conditional building is removed to favor simplicity, resulting in only one image with discourse and all the needed plugins being generated.


amandahla (dismissed):
There are some warnings about apt like "-no-install-recommends" and delete apt-get lists. Do you think would make a difference in the image size?

discourse.Dockerfile
```diff
@@ -29,97 +24,92 @@ ENV CONTAINER_APP_GID ${CONTAINER_APP_GID:-200}
...
-FROM ${IMAGE_TYPE} AS final
-# Redeclare IMAGE_TYPE variable so it can be referenced inside this build
-# stage per
-# https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
-ARG IMAGE_TYPE
-RUN echo "Used ${IMAGE_TYPE} image type"
+    && su -s /bin/bash -c 'gem install bundler' "${CONTAINER_APP_USERNAME}" \
+    && su -s /bin/bash -c '${CONTAINER_APP_ROOT}/app/bin/bundle install --gemfile=${PLUGINS_DIR}/discourse-saml/Gemfile --path=${PLUGINS_DIR}/discourse-saml/gems' "${CONTAINER_APP_USERNAME}" \
+    && ln -s "${PLUGINS_DIR}/discourse-saml/gems/ruby/"* "${PLUGINS_DIR}/discourse-saml/gems/" \
+    && su -s /bin/bash -c '${CONTAINER_APP_ROOT}/app/bin/bundle install' "${CONTAINER_APP_USERNAME}"
```
jdkandersson, line 115: Does this help with reducing the startup of the discourse charm? I have been using the charm and after the pod starts it has to run another ~5 minutes of scripts for it to be ready to serve. Not sure if it is possible to run those scripts in the Dockerfile rather than after startup

jdkandersson (dismissed):

discourse.Dockerfile
arturo-seijas: I plan to move some things in the script to the dockerfile. The leader election mechanism will change too, so that should also speed up things.

arturo-seijas:
> There are some warnings about apt like "-no-install-recommends" and delete apt-get lists. Do you think would make a difference in the image size?

Probably it will. The `--no-install-recommends`, however, won't have any effect since there's an `autoremove` later in the code. It would be good to change this but I'm leaving it for a future PR.

discourse.Dockerfile
mthaddon: Some of the start up time may be unavoidable as I understand it. Discourse needs to recompile all assets if any of its configuration variables have changed I believe, which obviously they will at run time versus build time.

arturo-seijas: That's my understanding too, but I think there's some for significant improvement.

gregory-schiano (approved):
LGTM

mthaddon (approved):
LGTM

discourse.Dockerfile
jdkandersson: :( Hopefully we can at least do something to improve startup time

arturo-seijas: That would be nice :)

PR #11, opened 2022-11-17
Drop support for database configurations

arturo-seijas:
Drop support for database configurations:
* Postgresql dabatase name override
* Redis database not provided through a relation
These features don't actually provide any advantage. From now on, the database name will always be the previous default and redis will have to be deployed in a charm.


mthaddon (approved):
LGTM, thx

gtrkiller (approved):
LGTM

PR #12, opened 2022-11-17
Drop support for ingress specific configurations

arturo-seijas:
Drop support for configuration of the following ingress options via the charm:
* TLS secret name
* Max body size of the request
* These configurations can be set directly through `juju config` in the related ingress charm, providing a more standard way to do so.


gtrkiller (approved):
LGTM

mthaddon (approved):
LGTM

PR #15, opened 2022-11-18
Set default hostname to application name

arturo-seijas:
Set default hostname to application name when not provided through configuration

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -13,7 +13,7 @@ options:
     default: true
   external_hostname:
     type: string
-    description: "External hostname this discourse instance should respond to. Defaults to the deployed application name."
+    description: "External hostname this discourse instance responds to. Defaults to [application name].local."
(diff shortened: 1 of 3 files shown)


config.yaml
```diff
@@ -13,7 +13,7 @@ options:
     default: true
   external_hostname:
     type: string
-    description: "External hostname this discourse instance should respond to. Defaults to the deployed application name."
```
mthaddon, line 16: I'm not sure why we're changing this to include a `.local` prefix? We do that for indico because the application won't respond unless there's a "." in the hostname, but I don't think that's the case for discourse is it?

arturo-seijas: Indeed. Let's do that

mthaddon (approved):
LGTM

weiiwang01 (approved):

PR #17, opened 2022-11-18
Remove locking mechanism

arturo-seijas:
Remove locking mechanism on the startup script.

There are three scenarios in which the script is executed:
* S3 configuration change: The migrations have already been run. Since leader election is guaranteed during 20s, there won't be concurrent executions of the migration.
* Installation: During installation, clashes might happen. Under this scenario the second pod that tries to run the migration will fail and retry later, given that no `BlockedStatus` is reached anymore on failure.
* Upgrade: charms are deployed as a statefulset with `maxUnavailable`1, so no concurrent executions will happen.


src/charm.py
```diff
@@ -341,48 +338,32 @@ def _config_changed(self, event):
...
-                self.model.unit.status = MaintenanceStatus(f"Executing setup {script}")
-                stdout, _ = process.wait_output()
-                logger.debug("%s stdout: %s", script, stdout)
-            except ExecError as e:
-                logger.error("%s command exited with code %d. Stderr:", script, e.exit_code)
-                for line in e.stderr.splitlines():
-                    logger.error("    %s", line)
-                logger.debug("%s stdout: %s", script, e.stdout)
-                self.model.unit.status = BlockedStatus(f"Error while executing {script}")
-                raise
```
gregory-schiano, line 365: I has been very very helpful to have the logs of the script when working with S3, I also used it a lot when debugging the integration tests on Github. I don't think it's a good idea to remove this part, unless you know a better way to get the script output

arturo-seijas: I plan to. This is still a draft and not even properly tested

src/charm.py
```diff
@@ -71,10 +71,8 @@ def __init__(self, *args):
             redis_relation={},
         )
         self.ingress = IngressRequires(self, self._make_ingress_config())
-        self.framework.observe(self.on.leader_elected, self._config_changed)
         self.framework.observe(self.on.discourse_pebble_ready, self._config_changed)
         self.framework.observe(self.on.config_changed, self._config_changed)
-        self.framework.observe(self.on.upgrade_charm, self._config_changed)
```
gregory-schiano, line 77: I think this is needed as when the charm gets updated, the application could be updated too and the pod_setup needs to run in order to proceed with the DB migrations

src/charm.py
```diff
@@ -341,48 +339,41 @@ def _config_changed(self, event):
...
             try:
-                self.model.unit.status = MaintenanceStatus(f"Executing setup {script}")
                 stdout, _ = process.wait_output()
                 logger.debug("%s stdout: %s", script, stdout)
             except ExecError as e:
                 logger.error("%s command exited with code %d. Stderr:", script, e.exit_code)
                 for line in e.stderr.splitlines():
                     logger.error("    %s", line)
-                logger.debug("%s stdout: %s", script, e.stdout)
-                self.model.unit.status = BlockedStatus(f"Error while executing {script}")
```
gregory-schiano, line 364: Will the charm put in BlockedState itself in case of Exception ?
Is so, what will be the message shown ? The idea here was to have a self-explaining message in case of BlockedState

gregory-schiano (approved):
LGTM

weiiwang01 (approved):

axinojolais:
> There are three scenarios in which the script is executed:
> 
>     * S3 configuration change: The migrations have already been run. Since leader election is guaranteed during 20s, there won't be concurrent executions of the migration.
> 
>     * Installation: During installation, clashes might happen. Under this scenario the second pod that tries to run the migration will fail and retry later, given that no `BlockedStatus` is reached anymore on failure.
> 
>     * Upgrade: charms are deployed as a statefulset with `maxUnavailable`1, so no concurrent executions will happen.

Looking back at why the lock got added, we also hit issues when scaling up the app. Is this something you've tested ?

arturo-seijas:
> > There are three scenarios in which the script is executed:
> > ```
> > * S3 configuration change: The migrations have already been run. Since leader election is guaranteed during 20s, there won't be concurrent executions of the migration.
> > 
> > * Installation: During installation, clashes might happen. Under this scenario the second pod that tries to run the migration will fail and retry later, given that no `BlockedStatus` is reached anymore on failure.
> > 
> > * Upgrade: charms are deployed as a statefulset with `maxUnavailable`1, so no concurrent executions will happen.
> > ```
> 
> Looking back at why the lock got added, we also hit issues when scaling up the app. Is this something you've tested ?

Yes, I have tested that too and no issues so far

PR #18, opened 2022-11-23
Remove default smtp address

arturo-seijas:
Remove default smtp address configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,9 @@ To deploy the charm and relate it to
     juju relate discourse-k8s postgresql-k8s:db-admin
     juju relate discourse-k8s redis-k8s
 
+The following config options are required in order for the charm to work: 
+`developer_emails`, `smtp_address` and `smtp_domain`.
+
 The charm also supports the `ingress` relation, which can be used with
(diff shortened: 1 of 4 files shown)


mthaddon:
What's the effect of this being empty? Might be worth mentioning that in the description in `config.yaml`.

weiiwang01 (dismissed):

weiiwang01 (dismissed):

mthaddon:
> > What's the effect of this being empty? Might be worth mentioning that in the description in `config.yaml`.
> 
> The charm will reach BlockedStatus. Analogous to what happens when the `smtp_domain` config is not specified

So won't this mean that the default deployment requires us to include this option now to work in any way? We generally prefer to have charms that can deploy clean (for local testing only) without having to set any config options.

mthaddon (changes requested):
Marking as needs changes because this would require anyone deploying this charm to set this option for even a local deploy for the charm to work in any way, which we'd like to avoid. If for some reason that's completely unavoidable, we should be updating the deployment instructions in the README as part of this change (and mentioning that the config option is required in`config.yaml`.

mthaddon (approved):
LGTM, thx

weiiwang01 (approved):

PR #20, opened 2022-12-06
COS integration

arturo-seijas:


weiiwang01 (dismissed):

image/scripts/app_launch
```diff
@@ -1,19 +1,11 @@
...
-cd "${CONTAINER_APP_ROOT}/app" || exit 1
-
 export UNICORN_BIND_ALL=0.0.0.0
 export UNICORN_SIDEKIQS=1
 
-# Putting the tail's in the background gets the various logs outputting to
-# stdout. The sed allows us to distinguish where a given log line comes form.
-tail -f "${CONTAINER_APP_ROOT}/app/log/production.log" | sed 's/^/discourse: /' &
-tail -f "${CONTAINER_APP_ROOT}/app/log/unicorn-stderr.log" | sed 's/^/unicorn-error: /' 1>&2 &
+su -s /bin/bash -c "${CONTAINER_APP_ROOT}/app/bin/bundle exec prometheus_exporter" &
```
amandahla, line 6: The log won't be redirected anymore to stdout?

amandahla (commented):
In the case of COS integration, is necessary to update the README.md?

image/scripts/app_launch
arturo-seijas: No, it will be collected by loki

arturo-seijas:
> In the case of COS integration, is necessary to update the README.md?

No, although me might want to expand the current docs

image/scripts/app_launch
```diff
@@ -1,19 +1,11 @@
...
+su -s /bin/bash -c "${CONTAINER_APP_ROOT}/app/bin/unicorn -c ${CONTAINER_APP_ROOT}/app/config/unicorn.conf.rb" "${CONTAINER_APP_USERNAME}" &
 
-# Add the discourse bin and scripts directories to the PATH.
-export PATH="${CONTAINER_APP_ROOT}/app/bin:${CONTAINER_APP_ROOT}/app/scripts:$PATH"
-su -s /bin/bash -c "${CONTAINER_APP_ROOT}/app/bin/unicorn -c ${CONTAINER_APP_ROOT}/app/config/unicorn.conf.rb" "${CONTAINER_APP_USERNAME}" 2>&1
+# If one of the processes exits, the other one will be killed so that the pod will be restarted by the failing probes
+wait -n
+kill "$(jobs -p)"
```
gregory-schiano, line 11: No fan of this coompared to the "trap" method, but that'll do the job
Any reasoning compared to the "trap" ?

src/charm.py
```diff
@@ -280,11 +315,20 @@ def _create_layer_config(self):
         }
         return layer_config
 
-    def _should_run_setup(self, current_plan, s3info: S3Info):
-        # Return true if no services are planned yet (first run)
-        return not current_plan.services or (
+    def _should_run_setup(self, current_plan: Dict, s3info: Optional[S3Info]) -> bool:
+        """Check if the setup script is to be run.
```
gregory-schiano, line 319: Perhaps elaborate a bit on the conditions that should be met ?

src/charm.py
```diff
@@ -83,8 +88,20 @@ def __init__(self, *args):
         self.redis = RedisRequires(self, self._stored)
         self.framework.observe(self.on.redis_relation_updated, self._config_changed)
 
-    def _make_ingress_config(self):
-        """Return a dict of our ingress config."""
+        self._metrics_endpoint = MetricsEndpointProvider(
+            self, jobs=[{"static_configs": [{"targets": ["*:9394"]}]}]
```
gregory-schiano, line 92: This port sounds like a magic number, is there a way to specify the port to the exporter service and therefore set the port to a constant value to ensure it'll always match ?

arturo-seijas: That would mean some additional complexity. Passing that value to the script and the to the appropriate command, instead I decided to rely on the default configuration.

src/charm.py
arturo-seijas: how about now?

image/scripts/app_launch
gregory-schiano: `trap 'trap - SIGTERM && kill 0' SIGINT SIGTERM EXIT`

This line and its comment you removed from that file

src/charm.py
```diff
@@ -83,8 +88,20 @@ def __init__(self, *args):
...
+        )
+        self._logging = LogProxyConsumer(
+            self, relation_name="logging", log_files=LOG_PATHS, container_name="discourse"
+        )
+        self._grafana_dashboards = GrafanaDashboardProvider(self)
+
+    def _make_ingress_config(self) -> Dict:
```
jdkandersson, line 99: It is possible to define the key and value type, although it is not always useful, using `Dict[str, str]`. If it gets too complex, it is also possible to use TypedDict: https://peps.python.org/pep-0589/

src/charm.py
```diff
@@ -122,18 +142,25 @@ def _is_config_valid(self) -> bool:
             errors.append("'saml_sync_groups' cannot be specified without a 'saml_target_url'")
 
         if self.config.get("s3_enabled"):
-            [
-                errors.append(f"'s3_enabled' requires '{s3_config}'")
-                for s3_config in REQUIRED_S3_SETTINGS
-                if not self.config[s3_config]
-            ]
+            errors.extend(
+                [
```
jdkandersson, line 146: Would it be worth making this a generator? It should just require changing the brackets from `[` to `(`, that way the list isn't created and then absorbed into `errors` and instead the elements are put straight into the list. Something like the following should work:

```Python
errors.extend(f"'s3_enabled' requires '{s3_config}'" for s3_config in REQUIRED_S3_SETTINGS if not self.config[s3_config)
```

src/charm.py
```diff
@@ -197,14 +227,19 @@ def _get_s3_env(self):
...
+            Dictionary with all the environment settings.
+        """
         # Get redis connection information from the relation.
         redis_hostname = None
         redis_port = 6379
-        for redis_unit in self._stored.redis_relation:
-            redis_hostname = self._stored.redis_relation[redis_unit]["hostname"]
-            redis_port = self._stored.redis_relation[redis_unit]["port"]
+        # This is the current recommended way of accessing the relation data.
+        for redis_unit in self._stored.redis_relation:  # type: ignore
```
jdkandersson, line 240: Out of curiosity, what is the mypy issue here?

src/charm.py
```diff
@@ -197,14 +227,19 @@ def _get_s3_env(self):
...
+            redis_hostname = self._stored.redis_relation[redis_unit].get("hostname")  # type: ignore
+            redis_port = self._stored.redis_relation[redis_unit].get("port")  # type: ignore
             logger.debug(
                 "Got redis connection details from relation of %s:%s", redis_hostname, redis_port
             )
```
jdkandersson, line 245: I know this wasn't changed in this PR, it seems that the configuration uses the `redis_hostname` and `redis_port` for the last `redis_unit` in `self._stored.redis_relation` and does so implicitly by relying on changes to a variable in a loop, would it be worth reviewing that logic to see if it is still appropriate? It could be more explicit. Maybe better for another PR?

src/charm.py
```diff
@@ -240,23 +275,23 @@ def _create_discourse_environment_settings(self):
 
         saml_config = self._get_saml_config()
         for key in saml_config:
-            pod_config[key] = saml_config[key]
+            pod_config.update({key: saml_config[key]})
```
jdkandersson, line 278: what about `pod_config.update(self._get_saml_config())`, that would negate the need for the for loop

src/charm.py
```diff
@@ -240,23 +275,23 @@ def _create_discourse_environment_settings(self):
...
             pod_config.update(self._get_s3_env())
 
         # We only get valid throttle levels here, otherwise it would be caught
-        # by `check_for_config_problems`, so we can be sure this won"t raise a
-        # KeyError.
-        for key in THROTTLE_LEVELS[self.config["throttle_level"]]:
-            pod_config[key] = THROTTLE_LEVELS[self.config["throttle_level"]][key]
+        # by `check_for_config_problems`.
+        if THROTTLE_LEVELS.get(self.config["throttle_level"]):
+            pod_config.update(THROTTLE_LEVELS.get(self.config["throttle_level"]))  # type: ignore
```
jdkandersson, line 286: Similar question here, what is the type issue?

src/charm.py
```diff
@@ -280,11 +315,20 @@ def _create_layer_config(self):
...
+        """Determine if the setup script is to be run based on the current plan and the new S3 settings.
+
+        Args:
+            current_plan: Dictionary containing the current plan.
+            s3info: S3Info object containing the S3 configuration options.
+
+        Returns:
+            If no services are planned yet (first run) or S3 settings have changed.
+        """
+        return not current_plan.services or (  # type: ignore
```
jdkandersson, line 328: What is the typing issue here?

src/charm.py
```diff
@@ -294,22 +338,25 @@ def _should_run_setup(self, current_plan, s3info: S3Info):
         )
 
     def _are_db_relations_ready(self) -> bool:
-        if not self._stored.db_name:
+        """Check if the needed database relations are established.
+
+        Returns:
+            If the needed relations have been established.
+        """
+        if not self._stored.db_name:  # type: ignore
```
jdkandersson, line 346: Here as well? Usually it is worthwhile to include a comment saying why a issue reported by `mypy` is not correct, same comment applies for the other `type: ignore`

src/grafana_dashboards/discourse.json
```diff
@@ -0,0 +1,2077 @@
+{
```
jdkandersson, line 1: I'll skip reviewing this line by line, is this auto generated somehow?

jdkandersson (commented):
Would a CORS integration test be useful? Maybe it is there an I missed it?

src/charm.py
arturo-seijas: error: overloaded function has no attribute "__iter__" (not iterable)  [attr-defined]

src/grafana_dashboards/discourse.json
arturo-seijas: Not auto generated but based on an official template. I've changed the datasource, removed the id and changed the descriptions and links too.

src/charm.py
arturo-seijas: Done

src/charm.py
arturo-seijas: Done

src/charm.py
arturo-seijas: I'll try to be a little more specific in these cases.

src/charm.py
arturo-seijas: Will add some comments

src/charm.py
arturo-seijas: Will add some comments

src/charm.py
arturo-seijas: Done

src/charm.py
arturo-seijas: We have several unpinned dependencies and I'd say an issue like that is as unlikely to happen as any other dependency  change breaking discourse. I'd rather be consistent here

image/scripts/app_launch
gregory-schiano: The trap command kills all process with the same process group which should work in your case as both processes are part of the same process group and inherits the signal trap

Did you try ?

src/charm.py
gregory-schiano: Well still not convinced as it's still a magic number that could lead to a break of the feature for no apparent reason by building a new docker image (which doesn't go through the integration test pipeline).
But if you're sure then so be it

Just one question, how does the charm behave if nothing is listening on the configured port ?

merkata (dismissed):
Looks good to me as is, approving.

src/charm.py
arturo-seijas: It will try to scrape that port anyway. We have an alert that will be triggered if it fails to do so.

arturo-seijas:
> Would a CORS integration test be useful? Maybe it is there an I missed it?

We had this discussion for Indico and decided it was overkill. We would be testing that COS is able to read the configurations and import the defined alerts, which is a COS responsibility. I've added a prometheus integration test though

src/charm.py
arturo-seijas: I've added an integration test that should catch this, btw

image/scripts/app_launch
gregory-schiano: Ok got it

tests/integration/test_charm.py
```diff
@@ -53,12 +53,30 @@ async def test_discourse_up(ops_test: OpsTest, app: Application, requests_timeou
...
+@pytest.mark.abort_on_fail
+async def test_prom_exporter_is_up(app: Application):
+    """
+    arrange: given charm in its initial state
+    act: when the metrics endpoint is scraped
+    assert: the response is 200 (HTTP OK)
+    """
+    # Application actually does have units
+    indico_unit = app.units[0]  # type: ignore
+    cmd = "curl http://localhost:9394/metrics"
```
gregory-schiano, line 71: Perhaps use a constant here instead of hardcoding the port on both sides ?

src/charm.py
gregory-schiano: Closing this discussion, at least there's a feedback for the user when it can't connect.
The integration test is good, too thx

tests/integration/test_charm.py
arturo-seijas: Done

weiiwang01 (approved):

gregory-schiano (approved):
LGTM

PR #23, opened 2022-12-09
Add action for creating admin user

arturo-seijas:
Add action for creating admin user. The action requires emails and password as parameters.
Also removes the requirement of having set up the SMTP settings for the charm to reach active status

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -474,7 +474,7 @@ class DiscourseCharm(CharmBase):
             except ExecError as ex:
                 logger.error(ex)
                 event.fail(
-                    # Parameter validation erros are printed to stdout
+                    # Parameter validation errors are printed to stdout
                     f"Failed to create user with email {email}: {ex.stdout}"  # type: ignore
                 )
             event.set_results({"user": f"{email}"})
--- a/src/charm.py
+++ b/src/charm.py
@@ -13,7 +13,7 @@ from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer
 from charms.nginx_ingress_integrator.v0.ingress import IngressRequires
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from charms.redis_k8s.v0.redis import RedisRelationCharmEvents, RedisRequires
-from ops.charm import CharmBase, HookEvent
+from ops.charm import ActionEvent, CharmBase, HookEvent
 from ops.framework import StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
@@ -25,6 +25,7 @@ pgsql = ops.lib.use("pgsql", 1, "postgresql-charmers@lists.launchpad.net")
 S3Info = namedtuple("S3Info", ["enabled", "region", "bucket", "endpoint"])
 
 DATABASE_NAME = "discourse"
+DISCOURSE_PATH = "/srv/discourse/app"
 THROTTLE_LEVELS = {
     "none": {
         "DISCOURSE_MAX_REQS_PER_IP_MODE": "none",
@@ -85,6 +86,7 @@ class DiscourseCharm(CharmBase):
             self.db.on.database_relation_joined, self._on_database_relation_joined
         )
         self.framework.observe(self.db.on.master_changed, self._on_database_changed)
+        self.framework.observe(self.on.add_admin_user_action, self._on_add_admin_user_action)
 
         self.redis = RedisRequires(self, self._stored)
         self.framework.observe(self.on.redis_relation_updated, self._config_changed)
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -453,6 +449,36 @@ def _on_database_changed(self, event: pgsql.DatabaseChangedEvent) -> None:  # ty
...
+            )
+            try:
+                process.wait_output()
+            except ExecError as ex:
+                logger.error(ex)
+                event.fail(
+                    # Parameter validation errors are printed to stdout
+                    f"Failed to create user with email {email}: {ex.stdout}"  # type: ignore
+                )
+            event.set_results({"user": f"{email}"})
```
nrobinaubertin, line 480: Should we set this result if the action fails ?

arturo-seijas: Fixed. Thanks

src/charm.py
```diff
@@ -453,6 +449,36 @@ def _on_database_changed(self, event: pgsql.DatabaseChangedEvent) -> None:  # ty
...
+                ],
+                user="discourse",
+                working_dir=DISCOURSE_PATH,
+                environment=self._create_discourse_environment_settings(),
+            )
+            try:
+                process.wait_output()
+                event.set_results({"user": f"{email}"})
+            except ExecError as ex:
+                logger.error(ex)
```
gregory-schiano, line 476: Perhaps logging stderr here also, as the stacktrace doesn't necessarily contains useful information in the case of a pebble exec

src/charm.py
```diff
@@ -453,6 +449,36 @@ def _on_database_changed(self, event: pgsql.DatabaseChangedEvent) -> None:  # ty
...
+        """
+        email = event.params["email"]
+        password = event.params["password"]
+        container = self.unit.get_container("discourse")
+        if container.can_connect():
+            process = container.exec(
+                [
+                    "bash",
+                    "-c",
+                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",
```
gregory-schiano, line 466: Do you know what happens if a user with this email address already exists ?

src/charm.py
arturo-seijas: It does in this case

tests/unit/test_charm.py
```diff
@@ -107,17 +92,9 @@ def test_config_changed_when_no_cors(self):
         assert: it will get to blocked status waiting for it.
         """
         self.add_database_relations()
+        self.harness.update_config({"cors_origin": ""})
```
merkata, line 95: Is this related to the creation of the admin user?

tests/unit/test_charm.py
```diff
@@ -130,17 +107,9 @@ def test_config_changed_when_throttle_mode_invalid(self):
         assert: it will get to blocked status waiting for a valid value to be provided.
         """
         self.add_database_relations()
+        self.harness.update_config({"throttle_level": "Scream"})
```
merkata, line 110: :scream:

merkata (approved):
Looks good!

tests/unit/test_charm.py
arturo-seijas: As part of this PR we're also removing the mandatory SMTP settings as a requirement to reach active status

gregory-schiano (approved):
LGTM

tests/unit/test_charm.py
```diff
@@ -399,8 +344,36 @@ def test_db_relation(self):
             "database name should be set after relation joined",
         )
 
+    def test_add_admin_user(self):
```
jdkandersson, line 347: We have a standard for the test docstring, is it applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

tests/unit/test_charm.py
```diff
@@ -399,8 +344,36 @@ def test_db_relation(self):
...
+        self.harness.disable_hooks()
+        self.add_database_relations()
+
+        charm: DiscourseCharm = typing.cast(DiscourseCharm, self.harness.charm)
+        with patch.object(Container, "exec", return_value=MockExecProcess()) as exec_mock:
+            self.harness.container_pebble_ready("discourse")
+
+            email = "sample@email.com"
+            password = "somepassword"  # nosec
+            event = MagicMock()
```
jdkandersson, line 357: It is possible to supply a spec to Mocks using the `spec` parameter to ensure that all the interactions conform to a class or instance: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock

tests/unit/test_charm.py
```diff
@@ -399,8 +344,36 @@ def test_db_relation(self):
...
+                "email": email,
+                "password": password,
+            }
+            charm._on_add_admin_user_action(event)
+
+        exec_mock.assert_any_call(
+            [
+                "bash",
+                "-c",
+                "./bin/bundle exec rake admin:create <<< $'sample@email.com\nsomepassword\nsomepassword\nY'",
```
jdkandersson, line 368: Would it be worth changing this to an f-string with the email and password defined earlier? E.g., `f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'"`

PR #29, opened 2023-01-10
feat(chaos): Bump duration of experiments

merkata:
This PR addresses a timeout in the chaos step of the CI/CD pipeline where the discourse charm does not become healthy in the specified duration (default of 60 seconds). A local test confirms the charm becomes stable, so with this change, the total duration of the chaos experiments is increased, so that the test can (hopefully) pass.

With fixing the pipeline run, some additional steps were needed:

* there is now a cleanup of S3 related variables in the integration test for S3
* I had to add dependencies for the linting to pass
* Functions in the unit and integration tests are refactored in order for pylint to pass
* Some pylint issues are disabled, please have a look and supply feedback
* I added a trivy.yaml config fail to ensure trivy doesn't break the pipeline and created an issue to address trivy findings
* I added a workflow action to sync Github issues to our Jira (so that the issue above lands in Jira as well)

*NOTE* I've cherry picked commits based on scope and have created additional PRs for Jira, Trivy and linting, so once these are merged, this PR will be slimmer and will have only Chaos test related changes. Respective issues (linting, trivy) are fixed in the respective PRs. Thanks!


arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -13,3 +13,4 @@ jobs:
       chaos-enabled: true
       chaos-experiments: pod-delete
       chaos-app-label: app.kubernetes.io/name=discourse-k8s
+      chaos-duration: 360
```
gregory-schiano, line 16: I suppose you tested locally to define this value ?
It will probably depends on the perf of the machine running the test, did you add some "buffer" when setting this value ?

merkata: I did a basic test of bringing up a clean microk8s cluster, installing the discourse charm with required relations and deleting the discourse pod, observing what will happen (and when). This was the local test, so not using Litmus Chaos directly - I thought that wouldn't be needed, however the test fails again and I need to see how to reproduce it, perhaps via mimicking the CI/CD - cloning the repo with experiments, setting environment variables and observing what happens. It's a tedious process I did when initially troubleshooting Litmus for the Indico charm and hoped I could skip that part. Alas.

gregory-schiano: Strange that it failed after 244s, the error doesn't explicitely states it's a timeout, so the issue could be different

amandahla (dismissed):

arturo-seijas (dismissed):
LGTM

amandahla (dismissed):

weiiwang01 (dismissed):

jdkandersson (dismissed):

arturo-seijas (dismissed):
LGTM

gregory-schiano (dismissed):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -13,3 +13,7 @@ jobs:
       chaos-enabled: true
       chaos-experiments: pod-delete
       chaos-app-label: app.kubernetes.io/name=discourse-k8s
+      chaos-interval: 600
```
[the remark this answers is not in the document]
merkata: Ordered, thanks for spotting this!

jdkandersson (dismissed):

gregory-schiano:
Unittests are failing due to the absence of pylint in the tox env

arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM! 6!6!6!

jdkandersson (dismissed):
It keeps getting longer...

merkata:
> It keeps getting longer...

and unfortunately still wouldn't pass - on a local setup it passes in 9min, however in CI I'm not sure how much it would take :/

arturo-seijas (dismissed):
LGTM

gregory-schiano:
> > It keeps getting longer...
> 
> and unfortunately still wouldn't pass - on a local setup it passes in 9min, however in CI I'm not sure how much it would take :/

May I suggest to use the debugging shell and run from the debugging shell the experiment, it'll allow you to check what's hapenning, and what's taking that long

merkata:
> > > It keeps getting longer...
> > 
> > 
> > and unfortunately still wouldn't pass - on a local setup it passes in 9min, however in CI I'm not sure how much it would take :/
> 
> May I suggest to use the debugging shell and run from the debugging shell the experiment, it'll allow you to check what's hapenning, and what's taking that long

That's a great suggestion, I did that previously when looking at an issue (wrong parsing of time) and it doesn't provide much insight as to why Litmus is failing. The thing I will do is simulate the test in pytest with a test that kills the discourse pod and waits for it to be alive, capturing the state and logs from the pod, that is something I'm missing in Litmus. This is getting a real 'private eye' vibe.

gregory-schiano:
> 

well at least using the shell you can see if it's really the pod taking really long to reboot using kubectl / juju commands

merkata:
> > 
> 
> well at least using the shell you can see if it's really the pod taking really long to reboot using kubectl / juju commands

TIL about that - thanks a bunch!

src/charm.py
```diff
@@ -461,7 +463,7 @@ def _on_add_admin_user_action(self, event: ActionEvent) -> None:
                 [
                     "bash",
                     "-c",
-                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",
+                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",  # pylint: disable=line-too-long
```
nrobinaubertin, line 466: I'm wondering if this is a case for using `format()` instead of a f-string with a pylint disable

merkata: Good! Let me rework that, I don't like disabling errors when they can be fixed. Thanks!

merkata:
I'm converting this to draft due to the CI/CD issue that needs more attention - I can't seem to reproduce it locally with or without pytest, although I think I'm following all steps from the pipeline to the extent possible. Stay tuned.

gregory-schiano:
Few linting issues needs to be solved and I think we'll be good with this one, isn't it ?

merkata:
> Few linting issues needs to be solved and I think we'll be good with this one, isn't it ?

Well - I can address the linting issues, however these are not my main concern. I'll see what I can do here in terms of ensuring the integration tests pass and then the chaos tests as well. What I found out is that enabling S3 on the S3 integration test causes a subsequent migration to break - I think that is expected as we're using localstack and are just interested in object creation, however recompiling assets in the rake job would like to access actual objects (that's my understanding). I'm trying to do a cleanup and setting S3 to false in the app config settings during the S3 test and hope the chaos test will pass. I'm not sure at this point whether I should do something else to have Discourse know S3 should not be used (might need to unset an ENV variable, don't know yet whether set_config does that for me, might need to do a migrate_from_s3 rake task as described here https://meta.discourse.org/t/how-to-turn-off-s3-storage/160338/6). With the latter we need to think about testing actual migrations, how to enable that and how to test it - migrate to S3, migrate from S3, what to do when access to S3 is lost etc.

.github/workflows/integration_test.yaml
```diff
@@ -8,8 +8,10 @@ jobs:
...
-      extra-arguments: --s3-url http://172.17.0.1:4566
-      pre-run-script: localstack-installation.sh
+      chaos-app-label: app.kubernetes.io/name=discourse-k8s
       chaos-enabled: true
       chaos-experiments: pod-delete
-      chaos-app-label: app.kubernetes.io/name=discourse-k8s
+      chaos-duration: 180
+      extra-arguments: --s3-url http://172.17.0.1:4566
+      pre-run-script: localstack-installation.sh
+      trivy-image-config: "trivy.yaml"
```
merkata, line 17: Supplied a `trivy.yaml` config file like we do on Indico and there's an open issue about addressing the found vulnerabilities.

src/charm.py
```diff
@@ -81,11 +81,11 @@ def __init__(self, *args):
         self.framework.observe(self.on.discourse_pebble_ready, self._config_changed)
         self.framework.observe(self.on.config_changed, self._config_changed)
 
-        self.db = pgsql.PostgreSQLClient(self, "db")
+        self.db_client = pgsql.PostgreSQLClient(self, "db")
```
merkata, line 84: Snake case best practice.

tests/integration/conftest.py
```diff
@@ -15,20 +16,20 @@
 logger = logging.getLogger(__name__)
 
 
-@fixture(scope="module")
-def metadata():
+@fixture(scope="module", name="metadata")
+def fixture_metadata():
```
merkata, line 20: Address pylint issues with redeclaring a variable.

tests/integration/test_charm.py
```diff
@@ -107,27 +108,23 @@ async def test_setup_discourse(
...
-        "Content-Type": "application/x-www-form-urlencoded",
-    }
-
     logger.info("Submitting registration form")
 
     response = session.post(
         f"{discourse_url}/finish-installation/register",
-        headers=form_headers,
-        data=urlencode(form_fields),
+        headers={
```
merkata, line 115: Reducing local variables, addresses pylint issue.

tests/integration/test_charm.py
```diff
@@ -246,28 +240,15 @@ async def test_s3_conf(ops_test: OpsTest, app: Application, s3_url: str):
...
-    s3_bucket = "tests"
-    s3_region = "us-east-1"
-
-    # Parse URL to get the IP address and the port, and compose the required variables
-    parsed_s3_url = urlparse(s3_url)
-    s3_ip_address = parsed_s3_url.hostname
-    s3_endpoint = f"{parsed_s3_url.scheme}://{s3_domain}"
-    if parsed_s3_url:
-        s3_endpoint = f"{s3_endpoint}:{parsed_s3_url.port}"
+    s3_conf: Dict = generate_s3_config(s3_url)
```
merkata, line 243: Reducing local variables, addresses pylint issue.

tests/integration/test_charm.py
```diff
@@ -326,10 +307,48 @@ async def test_s3_conf(ops_test: OpsTest, app: Application, s3_url: str):
...
+    assert s3_conf["bucket"] in bucket_list
 
     # Check content has been uploaded in the bucket
-    response = s3_client.list_objects(Bucket=s3_bucket)
+    response = s3_client.list_objects(Bucket=s3_conf["bucket"])
     object_count = sum(1 for _ in response["Contents"])
 
     assert object_count > 0
+
+    # Cleanup
```
merkata, line 318: Cleanup so that subsequent tests (like chaos tests) are executed properly not against a mock S3 (and they would fail).

tests/integration/test_charm.py
```diff
@@ -326,10 +307,48 @@ async def test_s3_conf(ops_test: OpsTest, app: Application, s3_url: str):
...
+            "s3_access_key_id": "",
+            # Default localstack region
+            "s3_region": "",
+        }
+    )
+    assert ops_test.model
+    await ops_test.model.wait_for_idle(status="active")
+
+
+def generate_s3_config(s3_url: str) -> Dict:
```
merkata, line 335: Helper for reducing local variables.

tests/unit/test_charm.py
```diff
@@ -140,30 +141,31 @@ def test_config_changed_when_s3_and_no_bucket_invalid(self):
             BlockedStatus("'s3_enabled' requires 's3_bucket'"),
         )
 
-    def test_config_changed_when_valid_no_s3_backup_nor_cdn(self):
+    @patch.object(Container, "exec")
```
merkata, line 144: Substitutes a class with no methods and addresses a pylint issue in that way.

tox.ini
```diff
@@ -36,16 +36,21 @@ description = Check code against coding style standards
...
     flake8-docstrings
+    flake8-docstrings-complete
     flake8-copyright
     flake8-builtins
+    flake8-test-docs
     isort
     mypy
     ops
     ops-lib-pgsql
+    psycopg2-binary
```
merkata, line 51: Needed for linting.

trivy.yaml
```diff
@@ -0,0 +1,4 @@
+exit-code: 0
```
merkata, line 1: Same config as the one in Indico charm repo. We need to discuss in the future whether we drive trivy configuration here or in CI/CD, as both don't work well. For instance in CI/CD trivy would filter out anything not WARNING or CRITICAL, but that behavior is overridden here and trivy would also report HIGH, MINOR etc.

gregory-schiano:
I'm not a big fan of leaving so many disabled line-too-long, wasn't there a better approach for these ?

merkata:
> I'm not a big fan of leaving so many disabled line-too-long, wasn't there a better approach for these ?

I did try finding a middle ground between black and pylint, did not succeed. To me it is trial and error - bend a line somewhere, see whether black bends it elsewhere, watch pylint complain again. There are 8 ignores, I'd leave this as an issue in the backlog with a `good-first-issue` label for someone to tackle. I'm too old for this.

.github/workflows/integration_test.yaml
```diff
@@ -9,11 +9,9 @@ jobs:
     secrets: inherit
     with:
       chaos-app-label: app.kubernetes.io/name=discourse-k8s
-      chaos-duration: 666
       chaos-enabled: true
       chaos-experiments: pod-delete
-      chaos-interval: 666
-      chaos-status-delay: 10
-      chaos-status-duration: 222
+      chaos-duration: 180
```
jdkandersson, line 14: Looks like we have a reduction in duration now, nice!

src/charm.py
```diff
@@ -162,7 +162,7 @@ def _get_saml_config(self) -> Dict[str, Any]:
             Dictionary with the SAML configuration settings..
         """
         saml_fingerprints = {
-            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"
+            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"  # pylint: disable=line-too-long
```
jdkandersson, line 165: I remember seeing something like this on another repo, the line length issue can be fixed without disabling the rule

src/charm.py
jdkandersson: It is also possible to break this line into multiple lines and retaining f-strings, see other PR that had similar issues

jdkandersson (commented):
This looks weirdly similar to another PR: https://github.com/canonical/discourse-k8s-operator/pull/33

merkata:
> This looks weirdly similar to another PR: #33

Yes, they are indeed similar - this PR here is the base one and I split it in 4 different PRs with cherry picks - one about Jira sync, one about Trivy config, one about linter issues and that one here related to the Chaos tests. This would be the last PR to merge and with syncing from main, only the chaos related changes will be here.

.github/workflows/integration_test.yaml
merkata: Yes, the underlying issue was elsewhere and this duration gets us to the desired result :)

merkata:
> I'm not a big fan of leaving so many disabled line-too-long, wasn't there a better approach for these ?

Well @jdkandersson is also not a big fan of line-too-long and left a bunch of great feedback on the feat/linter PR, so if you have a look at that, you'll see there are no more line-too-long disables :) Thanks for commenting!

src/charm.py
merkata: Resolved in feat/linters PR.

tests/integration/helpers.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+"""Helper functions for Discourse charm tests."""
```
arturo-seijas, line 2: Can you add the docstring after the license?

tests/integration/test_charm.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+"""Discourse integration tests."""
```
arturo-seijas, line 2: Same here

tests/unit/__init__.py
```diff
@@ -1,6 +1,7 @@
+"""Ops testing settings."""
```
arturo-seijas, line 1: Same here

tests/unit/test_charm.py
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env python3
+"""
```
arturo-seijas, line 2: Can you add the docstring after the license? Also, why a multiline comment

merkata: Hey sorry, I need to update the PR description - I split this PR into multiple ones (one for Jira integration, one for Trivy, one for addressing linting issues) and what you mention is addressed in the feat/linters PR. When it lands on main, I'll sync and this PR will have only chaos test related changes like it was supposed to be in the beginning :)

.github/workflows/issues.yaml
```diff
@@ -0,0 +1,11 @@
+name: Sync issues to Jira
+
+on:
+  issues:
+    # available via github.event.action
+    types: [opened, reopened, closed]
+
+jobs:
+  issues-to-jira:
+    uses: beliaev-maksim/github-to-jira-automation/.github/workflows/issues_to_jira.yaml@master
```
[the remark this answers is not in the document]
merkata: Exactly, good point.

gtrkiller (approved):
LGTM

amandahla (approved):
LGTM

jdkandersson:
58 comments and 46 commits later 😉 
Screenshot 2023-02-03 at 4 15 45 pm

PR #32, opened 2023-01-30
Feature/trivy conf

merkata:
This feature adds a trivy config file in the root directory `trivy.yaml` and references it in the integration tests action. That is similar to how it is done on Indico charm.

### DIFF
--- a/trivy.yaml
+++ b/trivy.yaml
@@ -1,0 +1,4 @@
+exit-code: 0
+timeout: 20m
+scan:
+  offline-scan: true
--- a/trivy.yaml
+++ b/trivy.yaml
@@ -1,4 +1,4 @@
-exit-code: 0
+exit-code: 2
 timeout: 20m
 scan:
   offline-scan: true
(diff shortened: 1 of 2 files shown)


arturo-seijas (dismissed):
LGTM

trivy.yaml
```diff
@@ -0,0 +1,4 @@
+exit-code: 2
```
gregory-schiano, line 1: What's the goal of returning 2 ?

gregory-schiano (commented):
Linting fails due to `SIMULATE_CAN_CONNECT` fix can be found on other projects

merkata:
> Linting fails due to `SIMULATE_CAN_CONNECT` fix can be found on other projects

That's fixed in a different PR, I broke up the chaos testing PR into several PRs and the fix is in the fix/linters branch.

trivy.yaml
merkata: Thanks for spotting, I cherry picked the wrong commit :) The `exit-code` needs to be `0`, I was testing earlier to see whether the config overrides what's in the github action for trivy. Thanks!

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

PR #33, opened 2023-01-30
Fix/linting

merkata:
This PR fixes linting issues found by mypy and pylint linters that are run via `tox -e lint`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,3 +50,6 @@ copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
 ignore_missing_imports = true
 explicit_package_bases = true
 namespace_packages = true
+
+[tool.pylint.SIMILARITIES]
+min-similarity-lines=10
--- a/src/charm.py
+++ b/src/charm.py
@@ -81,11 +81,11 @@ class DiscourseCharm(CharmBase):
         self.framework.observe(self.on.discourse_pebble_ready, self._config_changed)
         self.framework.observe(self.on.config_changed, self._config_changed)
 
-        self.db = pgsql.PostgreSQLClient(self, "db")
+        self.db_client = pgsql.PostgreSQLClient(self, "db")
         self.framework.observe(
-            self.db.on.database_relation_joined, self._on_database_relation_joined
+            self.db_client.on.database_relation_joined, self._on_database_relation_joined
         )
-        self.framework.observe(self.db.on.master_changed, self._on_database_changed)
+        self.framework.observe(self.db_client.on.master_changed, self._on_database_changed)
         self.framework.observe(self.on.add_admin_user_action, self._on_add_admin_user_action)
 
         self.redis = RedisRequires(self, self._stored)
@@ -162,7 +162,7 @@ class DiscourseCharm(CharmBase):
             Dictionary with the SAML configuration settings..
         """
         saml_fingerprints = {
-            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"
+            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"  # pylint: disable=line-too-long
         }
         saml_config = {}
 
@@ -317,7 +317,7 @@ class DiscourseCharm(CharmBase):
 
         Returns:
             If no services are planned yet (first run) or S3 settings have changed.
-        """
+        """  # pylint: disable=line-too-long
         # Properly type checks would require defining a complex TypedMap for the pebble plan
         return not current_plan.services or (  # type: ignore
             # Or S3 is enabled and one S3 parameter has changed
@@ -397,8 +397,8 @@ class DiscourseCharm(CharmBase):
             try:
                 stdout, _ = process.wait_output()
                 logger.debug("%s stdout: %s", script, stdout)
-            except ExecError as e:
-                logger.exception("%s command exited with code %d.", script, e.exit_code)
+            except ExecError as cmd_err:
+                logger.exception("%s command exited with code %d.", script, cmd_err.exit_code)
                 raise
 
         # Then start the service
@@ -461,7 +461,7 @@ class DiscourseCharm(CharmBase):
                 [
                     "bash",
                     "-c",
-                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",
+                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",  # pylint: disable=line-too-long
                 ],
                 user="discourse",
                 working_dir=DISCOURSE_PATH,
--- a/src/charm.py
+++ b/src/charm.py
@@ -162,7 +162,9 @@ class DiscourseCharm(CharmBase):
             Dictionary with the SAML configuration settings..
         """
         saml_fingerprints = {
-            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"  # pylint: disable=line-too-long
+            "https://login.ubuntu.com/+saml": (
+                "32:15:20:9F:A4:3C:8E:3E:8E:47:" "72:62:9A:86:8D:0E:E6:CF:45:D5"
+            )
         }
         saml_config = {}
 
@@ -309,7 +311,9 @@ class DiscourseCharm(CharmBase):
         return layer_config
 
     def _should_run_setup(self, current_plan: Dict, s3info: Optional[S3Info]) -> bool:
-        """Determine if the setup script is to be run based on the current plan and the new S3 settings.
+        """Determine if the setup script is to be run.
+
+           This is based on the current plan and the new S3 settings.
 
         Args:
             current_plan: Dictionary containing the current plan.
@@ -317,7 +321,7 @@ class DiscourseCharm(CharmBase):
 
         Returns:
             If no services are planned yet (first run) or S3 settings have changed.
-        """  # pylint: disable=line-too-long
+        """
         # Properly type checks would require defining a complex TypedMap for the pebble plan
         return not current_plan.services or (  # type: ignore
             # Or S3 is enabled and one S3 parameter has changed
@@ -461,7 +465,8 @@ class DiscourseCharm(CharmBase):
                 [
                     "bash",
                     "-c",
-                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",  # pylint: disable=line-too-long
+                    "./bin/bundle exec rake admin:create",
+                    f"<<< $'{email}\n{password}\n{password}\nY'",
                 ],
                 user="discourse",
                 working_dir=DISCOURSE_PATH,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,6 @@
-"""Module for test customizations."""
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
+"""Module for test customizations."""
 
 
 def pytest_addoption(parser):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,9 @@
+"""Module for test customizations."""
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 
 def pytest_addoption(parser):
+    """Adds parser switches."""
     parser.addoption("--discourse-image", action="store")
     parser.addoption("--s3-url", action="store")
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -15,20 +15,20 @@ from pytest_operator.plugin import OpsTest
 logger = logging.getLogger(__name__)
 
 
-@fixture(scope="module")
-def metadata():
+@fixture(scope="module", name="metadata")
+def fixture_metadata():
     """Provides charm metadata."""
     yield yaml.safe_load(Path("./metadata.yaml").read_text())
 
 
-@fixture(scope="module")
-def app_name(metadata):
+@fixture(scope="module", name="app_name")
+def fixture_app_name(metadata):
     """Provides app name from the metadata."""
     yield metadata["name"]
 
 
-@fixture(scope="module")
-def app_config():
+@fixture(scope="module", name="app_config")
+def fixture_app_config():
     """Provides app config."""
     yield {
         "developer_emails": "noreply@canonical.com",
@@ -76,7 +76,7 @@ async def app(ops_test: OpsTest, app_name: str, app_config: Dict[str, str], pyte
     await ops_test.model.wait_for_idle()
 
     # Add required relations
-    assert ops_test.model.applications[app_name].units[0].workload_status == WaitingStatus.name  # type: ignore
+    assert ops_test.model.applications[app_name].units[0].workload_status == WaitingStatus.name  # type: ignore # pylint: disable=line-too-long
     await asyncio.gather(
         ops_test.model.add_relation(app_name, "postgresql-k8s:db-admin"),
         ops_test.model.add_relation(app_name, "redis-k8s"),
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,3 +1,4 @@
+"""Discourse integration tests fixtures."""
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
@@ -18,7 +19,7 @@ logger = logging.getLogger(__name__)
 @fixture(scope="module", name="metadata")
 def fixture_metadata():
     """Provides charm metadata."""
-    yield yaml.safe_load(Path("./metadata.yaml").read_text())
+    yield yaml.safe_load(Path("./metadata.yaml").read_text(encoding="UTF-8"))
 
 
 @fixture(scope="module", name="app_name")
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -77,7 +77,8 @@ async def app(ops_test: OpsTest, app_name: str, app_config: Dict[str, str], pyte
     await ops_test.model.wait_for_idle()
 
     # Add required relations
-    assert ops_test.model.applications[app_name].units[0].workload_status == WaitingStatus.name  # type: ignore # pylint: disable=line-too-long
+    unit = ops_test.model.applications[app_name].units[0]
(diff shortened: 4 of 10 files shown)


tests/unit/__init__.py
```diff
@@ -1,6 +1,7 @@
+"""Ops testing settings."""
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 import ops.testing
 
-ops.testing.SIMULATE_CAN_CONNECT = True
+ops.testing.SIMULATE_CAN_CONNECT = True  # type: ignore[attr-defined]
```
arturo-seijas, line 7: You need to remove this. It's a breaking change in ops 2

pyproject.toml
```diff
@@ -50,3 +50,6 @@ copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
 ignore_missing_imports = true
 explicit_package_bases = true
 namespace_packages = true
+
+[tool.pylint.SIMILARITIES]
+min-similarity-lines=10
```
arturo-seijas, line 55: Why this change? Is it possible to refactor to avoid it?

arturo-seijas (changes requested):
I'm afraid you'll need to remove `ops.testing.CAN_CONNECT`instead of ignoring the linting error. This is the default behaviour in ops 2

pyproject.toml
merkata: Well it's a false-positive, pylint fails on this because there are several lines that are "found" in two files, the _patched_charm.py and test_charm.py, although in the first file the culprit is on line 61 with `SCRIPT_PATH = __import__("charm").SCRIPT_PATH` that seems to get expanded and pylint thinks it is a duplicate. I'm using this to reduce the noise. If there is another way I'd be happy to try, but this is what I ended up with.

tests/unit/__init__.py
merkata: Oh man, thanks for spotting! Will remove it.

tests/integration/helpers.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+"""Helper functions for Discourse charm tests."""
```
arturo-seijas, line 2: Shouldn't we keep the copyright notice before the docs?

tests/integration/test_charm.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+"""Discourse integration tests."""
```
arturo-seijas, line 2: Same here

tests/integration/helpers.py
merkata: Good point, fixed!

tests/integration/test_charm.py
merkata: Good point, fixed!

tests/integration/conftest.py
```diff
@@ -1,3 +1,4 @@
+"""Discourse integration tests fixtures."""
```
arturo-seijas, line 1: Same here

tests/unit/__init__.py
```diff
@@ -1,6 +1,3 @@
+"""Ops testing settings."""
```
arturo-seijas, line 1: Same here

tests/unit/test_charm.py
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env python3
+"""
```
arturo-seijas, line 2: Why is this not a single line comment instead of a block?

merkata: It's either me or `black`, but I bet it's me, will fix it :) Thanks!

tests/conftest.py
```diff
@@ -1,7 +1,9 @@
+"""Module for test customizations."""
```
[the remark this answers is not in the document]
merkata: Thanks, fixed!

tests/integration/conftest.py
merkata: Thanks, fixed!

tests/unit/__init__.py
merkata: Thanks, fixed!

arturo-seijas (dismissed):
LGTM

pyproject.toml
jdkandersson: The alternative is to disable the rule specifically on the offending line, that way we retain the linting rule on the other parts of the repository

src/charm.py
```diff
@@ -162,7 +162,7 @@ def _get_saml_config(self) -> Dict[str, Any]:
             Dictionary with the SAML configuration settings..
         """
         saml_fingerprints = {
-            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"
+            "https://login.ubuntu.com/+saml": "32:15:20:9F:A4:3C:8E:3E:8E:47:72:62:9A:86:8D:0E:E6:CF:45:D5"  # pylint: disable=line-too-long
```
jdkandersson, line 165: This can be broken into multiple lines, e.g.:

```Python
"https://login.ubuntu.com/+saml": (
    "32:15:20:9F:A4:3C:8E:3E:8E:47:"
    "72:62:9A:86:8D:0E:E6:CF:45:D5"
)
```

src/charm.py
```diff
@@ -317,7 +317,7 @@ def _should_run_setup(self, current_plan: Dict, s3info: Optional[S3Info]) -> boo
 
         Returns:
             If no services are planned yet (first run) or S3 settings have changed.
-        """
+        """  # pylint: disable=line-too-long
```
jdkandersson, line 320: I'm guessing the offending line is the first one, should we instead summarise it and add a more detailed description in the body? E.g.;

```Python
        """Determine if the setup script is to be run.

        This is based on the current plan and the new S3 settings.

        Args:
            current_plan: Dictionary containing the current plan.
            s3info: S3Info object containing the S3 configuration options.

         Returns:
             If no services are planned yet (first run) or S3 settings have changed.
         """
```

src/charm.py
```diff
@@ -461,7 +461,7 @@ def _on_add_admin_user_action(self, event: ActionEvent) -> None:
                 [
                     "bash",
                     "-c",
-                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",
+                    f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",  # pylint: disable=line-too-long
```
jdkandersson, line 464: Same comment here, this can be broken into multiple lines. black will do this automatically if you just insert two quotes where you want the line break to be

tests/integration/conftest.py
```diff
@@ -76,7 +77,7 @@ async def app(ops_test: OpsTest, app_name: str, app_config: Dict[str, str], pyte
     await ops_test.model.wait_for_idle()
 
     # Add required relations
-    assert ops_test.model.applications[app_name].units[0].workload_status == WaitingStatus.name  # type: ignore
+    assert ops_test.model.applications[app_name].units[0].workload_status == WaitingStatus.name  # type: ignore # pylint: disable=line-too-long
```
jdkandersson, line 80: This can also be resolved using:

```Python
unit = ops_test.model.applications[app_name].units[0]
assert unit.workload_status == WaitingStatus.name
```

tests/integration/test_charm.py
```diff
@@ -107,27 +108,23 @@ async def test_setup_discourse(
...
         f"{discourse_url}/finish-installation/register",
-        headers=form_headers,
-        data=urlencode(form_fields),
+        headers={
+            "Host": f"{app_config['external_hostname']}",
+            "Content-Type": "application/x-www-form-urlencoded",
+        },
+        data=urlencode(
+            {
+                "utf8": parsed_registration.body.find("input", attrs={"name": "utf8"}).get("value"),  # type: ignore # pylint: disable=line-too-long
```
jdkandersson, line 121: For type: ignore, should we add a comment on why? This is requested by our contributing guide: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis

jdkandersson, line 121: I think you can resolve this line to long as well using something like:

```Python
parsed_registration.body.find(
    "input", attrs={"name": "utf8"}
).get("value")  # type: ignore
```

Not sure where the type: ignore needs to go, just guessing

jdkandersson: This comment also applies to other disabling of line too long in this PR

jdkandersson: This also applies to any other type: ignore that were added

tests/unit/test_charm.py
```diff
@@ -345,46 +347,52 @@ def test_db_relation(self):
...
+        mock_exec.assert_any_call(
             [
                 "bash",
                 "-c",
                 f"./bin/bundle exec rake admin:create <<< $'{email}\n{password}\n{password}\nY'",
             ],
             user="discourse",
             working_dir=DISCOURSE_PATH,
-            environment=charm._create_discourse_environment_settings(),
+            environment=charm._create_discourse_environment_settings(),  # pylint: disable=protected-access
```
jdkandersson, line 382: If the `# pylint: disable=protected-access` is at the top of the file then it is disabled for the entire file. It is probably reasonable to do it for test files

tests/unit/test_charm.py
```diff
@@ -345,46 +347,52 @@ def test_db_relation(self):
...
-        self.harness.charm._stored.db_password = "somepasswd"  # nosec
-        self.harness.charm._stored.db_host = "dbhost"
-        self.db_relation_id = self.harness.add_relation("db", "postgresql")
+        self.harness.charm._stored.db_name = "discourse-k8s"  # pylint: disable=protected-access
+        self.harness.charm._stored.db_user = "someuser"  # pylint: disable=protected-access
+        self.harness.charm._stored.db_password = (  # pylint: disable=protected-access
+            "somepasswd"  # nosec
+        )
+        self.harness.charm._stored.db_host = "dbhost"  # pylint: disable=protected-access
+        self.db_relation_id = (  # pylint: disable=attribute-defined-outside-init
```
jdkandersson, line 393: Should we note why this is reasonable to do?

jdkandersson (commented):
Thanks for tackling the linting and typing issues!

src/charm.py
merkata: Black puts it back on one line like  "32:15:20:9F:A4:3C:8E:3E:8E:47:" "72:62:9A:86:8D:0E:E6:CF:45:D5", but it is still less than the max-line-length, so it works. Thanks!

src/charm.py
merkata: Definitely better, thanks!

tests/integration/conftest.py
merkata: Thanks, it is better that way!

tests/integration/test_charm.py
merkata: I think we should address that, however I haven't authored the ignores, they were part of other PRs, perhaps on a subsequent refactor?

merkata: Awesome, I tried something similar yesterday, but black reformatted it, however with your example it works! Thanks a lot!

tests/unit/test_charm.py
merkata: Good point! I placed it on top and added a comment. Thanks!

tests/unit/test_charm.py
merkata: Good point, noted!

arturo-seijas (dismissed):
LGTM

tests/integration/test_charm.py
```diff
@@ -212,13 +216,12 @@ async def test_setup_discourse(
...
     parsed_admin: BeautifulSoup = BeautifulSoup(response.content, features="html.parser")
     assert parsed_admin.head
-    csrf_token = parsed_admin.head.find("meta", attrs={"name": "csrf-token"}).get("content")  # type: ignore
+    csrf_token = parsed_admin.head.find("meta", attrs={"name": "csrf-token"}).get(  # type: ignore
+        "content"
+    )
 
     logger.info("Getting admin API key")
 
-    # Finally create an API Key, which will be used on the next integration tests
```
gregory-schiano, line 219: We should probably keep the comment

tests/integration/test_charm.py
```diff
@@ -212,13 +216,12 @@ async def test_setup_discourse(
     # Extract the CSRF token
     parsed_admin: BeautifulSoup = BeautifulSoup(response.content, features="html.parser")
     assert parsed_admin.head
-    csrf_token = parsed_admin.head.find("meta", attrs={"name": "csrf-token"}).get("content")  # type: ignore
+    csrf_token = parsed_admin.head.find("meta", attrs={"name": "csrf-token"}).get(  # type: ignore
+        "content"
+    )
```
gregory-schiano, line 221: Generic question, will black complain if we return before `.get(` ?
I don't find this kind of indentation very elegant and readable... but perhaps that's the python way

merkata: Good question - unfortunately black reformatted it this way, I set it differently, though without luck. If it wasn't for the type comment, it would fit in one line as well, perhaps that's confusing black in this case?

tests/integration/test_charm.py
merkata: We sure do, I don't know how it got deleted, I'm putting it back. Thanks!

arturo-seijas (approved):
LGTM

tests/integration/test_charm.py
gregory-schiano: Well this suggests to use `\`
https://stackoverflow.com/questions/8683178/chained-method-calls-indentation-style-in-python
=> Which I like for readability

But PEP8 specifies:
https://peps.python.org/pep-0008/#maximum-line-length
`The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.`

Which would suggest this:
```python
csrf_token = (parsed_admin.head.find("meta", attrs={"name": "csrf-token"}) # type: ignore
        .get( "content"))
```
Which I don't find very nice

@jdkandersson ?

src/charm.py
jdkandersson: So if you just write it like this: 

`"./bin/bundle exec rake admin:create <<< "f"$'{email}\n{password}\n{password}\nY'"`

`black` should automatically break this up for you

tests/integration/test_charm.py
jdkandersson: I'll leave that to you as a judgement call - it might be good to do it sooner rather than later to clean up the code base

tests/integration/test_charm.py
jdkandersson: I read it differently, it implies what `black` is doing by breaking on the opening bracket of `.get(`, generally speaking on formatting - whatever black does is right 😉

jdkandersson: note the `inside` in `Python’s implied line continuation inside parentheses`

jdkandersson: Depending on which part the `# type: ignore` applies I think you can trick black into breaking somewhere else, although I usually do this through trial and error. E.g., you could try this, I'm not sure what black would do with it though:

```Python
csrf_token = parsed_admin.head.find(  # type: ignore
    "meta", attrs={"name": "csrf-token"}).get("content"
)
```

jdkandersson: Anyway, the point of using black is so we don't need to debate formatting like this and just use whatever it does

merkata: > Depending on which part the `# type: ignore` applies I think you can trick black into breaking somewhere else, although I usually do this through trial and error. E.g., you could try this, I'm not sure what black would do with it though:
> 
> ```python
> csrf_token = parsed_admin.head.find(  # type: ignore
>     "meta", attrs={"name": "csrf-token"}
> ).get("content")
> ```

Using that example, black re-formats it to the above (what is already commited). At this point I'm fine with black doing its thing. I don't have more suggestions.

gregory-schiano: Well I don't find that black is doing something very readable in that case, having a single string and parenthesis on their own line look ugly to me.
```python
csrf_token = parsed_admin.head.find(  # type: ignore
    "meta", attrs={"name": "csrf-token"}
).get("content")
```

This formatting is way better and more readable. And for sure we can trust black to avoid formatting discussion, but these tools aren't the alpha and the omega... they're just using algorithm and could produce something that can be improved while still being valid :)

pyproject.toml
merkata: There is no offending line, placing the disable on some of the lines quoted in the error does not produce the desired result, even placing the disable on top of the file does not disable it, perhaps due to the import. More on the issue here https://github.com/PyCQA/pylint/issues/214 and also here https://github.com/snapcore/snapcraft/pull/3655

tests/integration/test_charm.py
merkata: I'm seeing such ignores piling up in other repos as well, let's have a discussion on the topic, we should address the root of why `mypy` is complaining and how to fix it. I'm dropping it until we figure out a holistic approach for all of our charm repos.

tests/integration/test_charm.py
merkata: Based on the feedback I don't know what to do next. I don't have approvals either, so is there a suggestion I can try and implement?

gregory-schiano: I'm in favor of the formatting I quoted in the previous comment

merkata: Yes, but

```shell
merkata@mastodon:~/dev/discourse-k8s-operator$ grep "csrf_token =" tests/integration/test_charm.py -A3
    csrf_token = parsed_admin.head.find(  # type: ignore
        "meta", attrs={"name": "csrf-token"}
    ).get("content")

merkata@mastodon:~/dev/discourse-k8s-operator$ black tests/integration/test_charm.py 
reformatted tests/integration/test_charm.py

All done! ✨ 🍰 ✨
1 file reformatted.
merkata@mastodon:~/dev/discourse-k8s-operator$ grep "csrf_token =" tests/integration/test_charm.py -A3
    csrf_token = parsed_admin.head.find("meta", attrs={"name": "csrf-token"}).get(  # type: ignore
        "content"
    )
```

amandahla (approved):
LGTM but it seems that trivy config was not found.

https://github.com/canonical/discourse-k8s-operator/actions/runs/4054111097/jobs/6975542344

Example of output in Indico:
```
"-a image" "-b table" "-c " "-d 1" "-e false" "-f os,library" "-g CRITICAL,HIGH" "-h " "-i ghcr.io/canonical/indico:4073362325" "-j ." "-k " "-l " "-m " "-n " "-o " "-p " "-q " "-r false" "-s " "-t " "-u " "-v trivy.yaml" "-z "
Running Trivy with trivy.yaml config from:  trivy.yaml
```

PR #34, opened 2023-02-02
fix(jira): Point to reusable workflow

merkata:
This fix replaces the original action to sync Github issues to Jira with the one under `operator-workflows` that is created as a reusable action. The original action has to be copied in the repo that should make use of it, so it made sense to have it in operator-workflows and charm repos to consume it.


jdkandersson (approved):

arturo-seijas (approved):
LGTM

PR #36, opened 2023-02-23
Add an integration test for SAML authentication

weiiwang01:


actions.yaml
```diff
@@ -8,4 +8,10 @@ add-admin-user:
       type: string
       description: User password.
   required: [email, password]
+force-https:
+  description: Configure Discourse to use https. Use lowercase true/false word values.
+  params:
+    force_bool:
+      type: string
+      default: "true"
```
mthaddon, line 16: I'm not sure we should be doing that via an action. Ideally we'd do this based on how the ingress is configured... If that's not going to be possible for now we may consider another approach though. I think the ingress relation passes has the end URL? Could we use that to confirm if we should force https rather then requiring it to be an action?

config.yaml
```diff
@@ -26,7 +26,7 @@ options:
   saml_target_url:
     type: string
     description: "SAML authentication target url."
-    default: ""
+    default: "https://login.staging.ubuntu.com/+saml"
```
mthaddon, line 29: This shouldn't be the default, let's leave the default as "".

.github/workflows/integration_test_with_secrets.yaml
```diff
@@ -0,0 +1,38 @@
...
+          docker build -f discourse.Dockerfile -t localhost:32000/discourse:test .
+          docker push localhost:32000/discourse:test
+
+      - name: Run integration tests
+        run: |
+          tox -e integration -- \
+          -m requires_secrets \
+          --saml-email ${{ secrets.TEST_SAML_EMAIL }} \
+          --saml-password ${{ secrets.TEST_SAML_PASSWORD }} \
+          --discourse-image=localhost:32000/discourse:test
```
arturo-seijas, line 38: Missing blank line

actions.yaml
weiiwang01: The ingress relation url will always be HTTP, also we will drop the ingress relation very soon.

```python
            event.relation.data[self.model.app]["url"] = f"http://{name}:{port}/"
```

I agree this shouldn't be a action, maybe this should be a configuration?

tests/unit/__init__.py
```diff
@@ -1,3 +1,8 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
+
 """Ops testing settings."""
+
+import ops.testing
+
+ops.testing.SIMULATE_CAN_CONNECT = True  # type: ignore
```
yanksyoon, line 8: Out of curiosity, what is the intention behind `SIMULATE_CAN_CONNECT`? I've removed in WordPress charm on this [commit](https://github.com/canonical/wordpress-k8s-operator/commit/93d21d5dd00c39ad38c6ff0fa8f2bf79ba16ad87) before, was wondering in which test this is required. :)

tests/unit/_patched_charm.py
```diff
@@ -34,7 +34,7 @@ class _PGSQLPatch:
     def __init__(self):
         # borrow some code from
         # https://github.com/canonical/ops-lib-pgsql/blob/master/tests/test_client.py
-        self._leadership_data = {}
+        self._leadership_data = {}  # type: ignore
```
yanksyoon, line 37:
```suggestion
        self._leadership_data:dict = {}
```
would this work here?

tests/unit/test_charm.py
```diff
@@ -1,21 +1,27 @@
...
 
+import ops.model
+import ops.pebble
 from ops.charm import ActionEvent
 from ops.model import ActiveStatus, BlockedStatus, Container, WaitingStatus
 from ops.testing import Harness
 
 from tests.unit._patched_charm import DISCOURSE_PATH, SCRIPT_PATH, DiscourseCharm, pgsql_patch
 
+# Copyright 2022 Canonical Ltd.
```
yanksyoon, line 22:
```suggestion
# Copyright 2023 Canonical Ltd.
```

.github/workflows/integration_test_with_secrets.yaml
```diff
@@ -0,0 +1,38 @@
+name: Integration tests (require secrets)
```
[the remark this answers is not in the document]
weiiwang01: This can not be done in the `operator-workflow`, secrets can not be passed through the workflow input. This is a limitation of GitHub action.

tox.ini
```diff
@@ -109,7 +109,7 @@ description = Run integration tests
 deps =
     boto3
     bs4
-    juju
+    juju==3.0.4
```
yanksyoon, line 112:
```suggestion
    juju>=3.0.4
```
Correct me if i've misunderstood, can this work like so? So that future versions can also work.

yanksyoon (dismissed):
LGTM! Minor suggestions :) Thank you

tests/unit/__init__.py
arturo-seijas: Iirc it has been deprecated

actions.yaml
mthaddon: Adding a new config option to enable this sounds good to me.

tests/unit/_patched_charm.py
weiiwang01: Not tested, but I think it will need both the key type and value type.

tests/unit/test_charm.py
weiiwang01: Fixed, thanks!

tox.ini
weiiwang01: Python juju client > 3.0.4 has some compatibility problems, so pinning the Python juju client to 3.0.4 for now.

.github/workflows/integration_test_with_secrets.yaml
weiiwang01: Fixed, thanks!

config.yaml
weiiwang01: Fixed, thanks!

actions.yaml
weiiwang01: A new config `force_https` has been added and the `force-https` action was removed, thanks!

docs/explanation/charm-architecture.md
```diff
@@ -72,6 +72,7 @@ For this charm, the following events are observed:
 
 1. [<container name>_pebble_ready](https://juju.is/docs/sdk/container-name-pebble-ready-event): fired on Kubernetes charms when the requested container is ready. Action: wait for the integrations, and configure the containers.
 2. [config_changed](https://juju.is/docs/sdk/config-changed-event): usually fired in response to a configuration change using the GUI or CLI. Action: wait for the integrations, validate the configuration, update Ingress, and restart the containers.
+4. [force_https_action](https://charmhub.io/discourse-k8s/actions): fired when force-https action is executed. Action: Force discourse to use https instead of http.
```
gtrkiller, line 75: If we removed the force-https action this should be modified too

tests/unit/test_charm.py
```diff
@@ -114,7 +153,7 @@ def test_config_changed_when_throttle_mode_invalid(self):
 
         self.assertEqual(
             self.harness.model.unit.status.name,
-            BlockedStatus.name,
+            BlockedStatus.name,  # type: ignore
```
amandahla, line 156: @gtrkiller did something like this in a similar [case](https://github.com/canonical/nginx-ingress-integrator-operator/pull/53):

```
# Mype can't recognize the name as a string type, so we should skip the type check.
ACTIVE_STATUS_NAME = ActiveStatus.name  # type: ignore[has-type]
...
 assert app_status == ACTIVE_STATUS_NAME
```

weiiwang01: Okay, I defined the value of `BlovkedStatus.name` as a global variable.

tests/integration/conftest.py
```diff
@@ -54,8 +54,34 @@ def requests_timeout():
     yield 15
 
 
-@pytest_asyncio.fixture(scope="module")
-async def app(ops_test: OpsTest, app_name: str, app_config: Dict[str, str], pytestconfig: Config):
+@fixture(scope="module")
+def run_action(ops_test: OpsTest) -> Callable[..., Awaitable[Any]]:
```
jdkandersson, line 58: What is the reason for making this a fixture please? Would this be better coded as a helper function? It seems quite helpful

tests/integration/conftest.py
```diff
@@ -82,7 +109,20 @@ async def app(ops_test: OpsTest, app_name: str, app_config: Dict[str, str], pyte
...
+@pytest_asyncio.fixture(scope="module")
+async def setup_saml_config(ops_test: OpsTest, app: Application):
+    """Set SAML related charm config to enable SAML authentication."""
+    assert ops_test.model
+    discourse_app = ops_test.model.applications[app.name]
+    await discourse_app.set_config(
+        {"saml_target_url": "https://login.staging.ubuntu.com/+saml", "force_https": "true"}
+    )
+    yield
+    await discourse_app.set_config({"saml_target_url": "", "force_https": "false"})
```
jdkandersson, line 128: Should this revert back to the starting configuration rather than hardcode? E.g., capture the configuration at the start of the fixture, set the configuration and then revert back to the starting configuration?

tests/integration/test_charm.py
```diff
@@ -361,3 +365,107 @@ def generate_s3_config(s3_url: str) -> Dict:
     s3_config["ip_address"] = s3_ip_address
     s3_config["endpoint"] = s3_endpoint
     return s3_config
+
+
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+@pytest.mark.requires_secrets
+@pytest.mark.usefixtures("setup_saml_config")
+async def test_saml_login(  # pylint: disable=too-many-locals
```
jdkandersson, line 374: Should this include a reason why the linting check is disabled?

tests/integration/test_charm.py
```diff
@@ -361,3 +365,107 @@ def generate_s3_config(s3_url: str) -> Dict:
...
+    pytestconfig: pytest.Config,
+    requests_timeout: int,
+    run_action,
+):
+    """
+    arrange: after discourse charm has been deployed, with all required relation established.
+    act: add an admin user and enable force-https mode.
+    assert: user can login discourse using SAML Authentication.
+    """
+    assert ops_test.model
```
jdkandersson, line 386: Would it be worth making this a fixture so the `assert ops_test.model` is only required once in that fixture?

tests/integration/test_charm.py
```diff
@@ -361,3 +365,107 @@ def generate_s3_config(s3_url: str) -> Dict:
...
+    urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+    email = pytestconfig.getoption("--saml-email")
```
jdkandersson, line 388: Should this be a fixture?

tests/integration/test_charm.py
```diff
@@ -361,3 +365,107 @@ def generate_s3_config(s3_url: str) -> Dict:
...
+    password = pytestconfig.getoption("--saml-password")
```
jdkandersson, line 389: Should this be a fixture?

tests/integration/test_charm.py
```diff
@@ -361,3 +365,107 @@ def generate_s3_config(s3_url: str) -> Dict:
...
+    username = email.split("@")[0]
+    host = app.name
+    original_getaddrinfo = socket.getaddrinfo
+
+    def patched_getaddrinfo(*args):
+        if args[0] == host:
+            return original_getaddrinfo("127.0.0.1", *args[1:])
+        return original_getaddrinfo(*args)
+
+    with unittest.mock.patch.multiple(socket, getaddrinfo=patched_getaddrinfo):
```
jdkandersson, line 408: What is the reason for this patch please?

tests/integration/test_charm.py
```diff
@@ -361,3 +365,107 @@ def generate_s3_config(s3_url: str) -> Dict:
...
+            verify=False,
+            timeout=requests_timeout,
+        )
+
+        preference_page = session.get(
+            f"https://{host}/u/{username}/preferences/account",
+            verify=False,
+            timeout=requests_timeout,
+        )
+        assert preference_page.status_code == 200
```
jdkandersson, line 471: Nice, this test checks that the user actually works!

tests/unit/_patched_charm.py
jdkandersson: I'm guessing mypy would not complain once the dict type is used

tests/integration/conftest.py
weiiwang01: The reason that this is not a helper function is that this function requires access to the `ops_test`, which is a fixture.

tests/integration/test_charm.py
weiiwang01: I don't think mypy can get context across fixtures.

tests/integration/test_charm.py
weiiwang01: This option is only used in this test case, put this in the fixture might complicated the error handling. In this test case, an error need to raise if the option is not provided, but for other test cases, missing such option is not an error. If keep the option extraction in the test case, I can raise an error message with the test case name in it. But if I put the option extraction in a fixture, the error message mentioning the required test cases name seems inappropriate.

tests/integration/test_charm.py
weiiwang01: Same as above.

tests/integration/conftest.py
weiiwang01: Updated, thanks!

tests/integration/test_charm.py
jdkandersson: What I was thinking was that the fixture would be something like this:

```python
@pytest.fixture
def model(ops_test):
    assert ops_test.model
    yield model
```

And then the test would be:

```python
def test_something(model: Model):
    ...
```

jdkandersson: So the fixture guarantees that the model is not `None` and then the type hint for the fixture can just be the model

tests/integration/test_charm.py
jdkandersson: I'm guessing that pytest only invokes fixture that are actually required for a test run. So, if a run doesn't need the fixture with this option, then it wouldn't get executed and no error is raised. The error message in the fixture could be "the invoked tests require the <option> to be able to run"

weiiwang01: Okay, I update the code to put SAML arguments into a fixture.

tests/integration/test_charm.py
weiiwang01: Updated, thanks!

gtrkiller (approved):
LGTM

mthaddon (approved):
LGTM, thx. I notice there are some pending comments, might be worth checking with the authors before merging, but otherwise great work!

tests/unit/_patched_charm.py
weiiwang01: Yes, but unfortunately, this is not the case. This is used internally in `pgsql` charm library.

tests/integration/test_charm.py
weiiwang01: Yes, but I think this line is pretty self-explanatory: there are too many local variables. This test case is quite long but pretty streamlined, so it wouldn't have a lot of readability issues.

weiiwang01:
> LGTM, thx. I notice there are some pending comments, might be worth checking with the authors before merging, but otherwise great work!

Everything resolved, merging now. Thanks!

PR #37, opened 2023-03-03
Add action to anonymize user data

arturo-seijas:
Add action to anonymize user data, ran as:
`juju run-action discourse-k8s/0 anonymize-user username=prueba --wait`


discourse.Dockerfile
```diff
@@ -68,9 +58,10 @@ RUN ln -s /usr/share/zoneinfo/UTC /etc/localtime \
 # Run build process.
     && git -C "${CONTAINER_APP_ROOT}" clone --depth 1 --branch "${CONTAINER_APP_VERSION}" https://github.com/discourse/discourse.git app
 
-# Apply patch for LP#1903695
+# Apply patches
 COPY image/patches /srv/patches
 RUN git -C "${CONTAINER_APP_ROOT}/app" apply /srv/patches/lp1903695.patch \
+    && git -C "${CONTAINER_APP_ROOT}/app" apply /srv/patches/anonymize_user.patch \
```
mthaddon, line 64: We should capture somewhere where/how we're trying to get this upstream, so that we can track if/when it lands and we can remove the custom patch.

actions.yaml
```diff
@@ -8,3 +8,10 @@ add-admin-user:
       type: string
       description: User password.
   required: [email, password]
+anonymize-user:
+  description: Anonymize a user.
+  params:
+    username:
+      type: string
+      description: Username.
```
gtrkiller, line 16: does it make sense to add a redundant description? (it is equal to the name of the field)

discourse.Dockerfile
arturo-seijas: This won't land in this Discourse version. We'll have to upgrade.

actions.yaml
arturo-seijas: Any suggestion?

src/charm.py
```diff
@@ -510,6 +506,35 @@ def _config_force_https(self) -> None:
         )
         process.wait_output()
 
+    def _on_anonymize_user_action(self, event: ActionEvent) -> None:
+        """Anonymize data from a user.
+
+        Args:
+            event: Event triggering the anonymize_admin_user action.
+        """
+        username = event.params["username"]
```
amandahla, line 515: Would be useful to check if the user is already anonymized?

Is it possible to do this even with administrator users?

actions.yaml
jdkandersson: Maybe `The unique identifier of the user to anonymize`?

image/patches/anonymize_user.patch
```diff
@@ -0,0 +1,16 @@
+--- a/lib/tasks/users.rake
++++ b/lib/tasks/users.rake
```
jdkandersson, line 2: Is there any other way to do this? We had something similar for Indico for creating an admin user. Is there a spec for this change?

src/charm.py
```diff
@@ -27,27 +27,26 @@
...
-    },
-    "strict": {
-        "DISCOURSE_MAX_REQS_PER_IP_MODE": "block",
-        "DISCOURSE_MAX_REQS_PER_IP_PER_MINUTE": 200,
-        "DISCOURSE_MAX_REQS_PER_IP_PER_10_SECONDS": 50,
-        "DISCOURSE_MAX_USER_API_REQS_PER_MINUTE": 100,
-        "DISCOURSE_MAX_ASSET_REQS_PER_IP_PER_10_SECONDS": 200,
-        "DISCOURSE_MAX_REQS_RATE_LIMIT_ON_PRIVATE": "false",
-    },
+THROTTLE_LEVELS: Dict = defaultdict(dict)
```
jdkandersson, line 30: I'm curios for the reason for this change?

src/charm.py
```diff
@@ -283,9 +283,7 @@ def _create_discourse_environment_settings(self) -> Dict[str, Any]:
 
         # We only get valid throttle levels here, otherwise it would be caught
         # by `check_for_config_problems`.
-        if THROTTLE_LEVELS.get(self.config["throttle_level"]):
-            # self.config return an Any type
```
jdkandersson, line 287: Should this comment explaining the `type: ignore` be retained?

src/charm.py
```diff
@@ -486,7 +483,6 @@ def _on_add_admin_user_action(self, event: ActionEvent) -> None:
                 process.wait_output()
                 event.set_results({"user": f"{email}"})
             except ExecError as ex:
-                logger.exception("Action add-admin-user failed")
```
jdkandersson, line 489: Is this log no longer needed?

src/charm.py
```diff
@@ -510,6 +506,35 @@ def _config_force_https(self) -> None:
         )
         process.wait_output()
 
+    def _on_anonymize_user_action(self, event: ActionEvent) -> None:
+        """Anonymize data from a user.
+
+        Args:
+            event: Event triggering the anonymize_admin_user action.
```
jdkandersson, line 513: is it only for admin users? The function name doesn't indicate that?

src/charm.py
```diff
@@ -510,6 +506,35 @@ def _config_force_https(self) -> None:
...
+                environment=self._create_discourse_environment_settings(),
+            )
+            try:
+                process.wait_output()
+                event.set_results({"user": f"{username}"})
+            except ExecError as ex:
+                event.fail(
+                    # Parameter validation errors are printed to stdout
+                    f"Failed to anonymize user with username {username}:"
+                    f"{ex.stdout}"  # type: ignore
```
jdkandersson, line 535: What is this `type: ignore` please? This standard requests a comment explaining the why to be added for any rule disables, is that applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis

tests/unit/test_charm.py
```diff
@@ -425,6 +425,37 @@ def test_add_admin_user(self, mock_exec):
             timeout=60,
         )
 
+    @patch.object(Container, "exec")
+    def test_anonymize_user(self, mock_exec):
+        """
+        arrange: set up discourse
+        act: execute the _on_anonymize_user_action method
+        assert: the underlying rake command to anonymize the user is executed
+        with the appropriate parameters.
```
jdkandersson, line 434: Should this indented by 1 level?

tests/unit/test_charm.py
```diff
@@ -425,6 +425,37 @@ def test_add_admin_user(self, mock_exec):
             timeout=60,
         )
 
+    @patch.object(Container, "exec")
+    def test_anonymize_user(self, mock_exec):
```
jdkandersson, line 429: Should this also have an integration test?

.github/load_tests/main.js
```diff
@@ -0,0 +1,16 @@
...
+  duration: '1m',
+  vus: 50,
+  thresholds: {
+    http_req_failed: ['rate<0.01'], // http errors should be less than 1%
+    http_req_duration: ['p(95)<500'], // 95 percent of response times must be below 500ms
+  },
+};
+
+export default function () {
+  http.get('http://indico.local');
```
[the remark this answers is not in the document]
arturo-seijas: Not sure how this file got here...

image/patches/anonymize_user.patch
arturo-seijas: It's possible to do it making a couple of API calls, one to retrieve the user id from the username and another one to anonymize, however, there are two reasons to prefer this approach:
* It is way simpler to leverage the rake tasks
* Although patching the code is far from ideal, this approach is already in place for fix https://github.com/discourse/discourse/pull/20523
* The patch is temporary, it will be removed when the https://github.com/discourse/discourse/pull/20522 is merged and Discourse, upgraded

src/charm.py
arturo-seijas: It allows us to remove one `if`

src/charm.py
arturo-seijas: Description fixed

src/charm.py
arturo-seijas: If it has, the anonymization will be applied again, having no effect whatsoever.

Yes, it is possible to anonymize admin usres.

src/charm.py
arturo-seijas: Added an explanation

tests/unit/test_charm.py
arturo-seijas: It seems overkill to me given criticallity

src/charm.py
arturo-seijas: Not a very helpful log entry

discourse.Dockerfile
```diff
@@ -68,9 +58,12 @@ RUN ln -s /usr/share/zoneinfo/UTC /etc/localtime \
 # Run build process.
     && git -C "${CONTAINER_APP_ROOT}" clone --depth 1 --branch "${CONTAINER_APP_VERSION}" https://github.com/discourse/discourse.git app
 
-# Apply patch for LP#1903695
+# Apply patches
```
mthaddon, line 61: Can we retain the reference to LP#1903695 here? I realise it's in the name of the patch too, but if we're listing references it'd be worth listing them all I think

mthaddon (approved):

amandahla (approved):
LGTM

tests/unit/test_charm.py
jdkandersson: Not so much about the patch, it is more about that the action being added works on a real version of the charm

arturo-seijas: It's unit tested so I'm assuming Discourse's contract is fine (once the changes land). It has been manually tested, oc.

discourse.Dockerfile
```diff
@@ -1,24 +1,14 @@
...
-ENV CONTAINER_APP_VERSION ${CONTAINER_APP_VERSION:-v2.7.10}
-ENV CONTAINER_APP_USERNAME ${CONTAINER_APP_USERNAME:-discourse}
-ENV CONTAINER_APP_UID ${CONTAINER_APP_UID:-200}
-ENV CONTAINER_APP_GROUP ${CONTAINER_APP_GROUP:-discourse}
-ENV CONTAINER_APP_GID ${CONTAINER_APP_GID:-200}
+ENV CONTAINER_APP_VERSION v2.7.10
+ENV CONTAINER_APP_USERNAME discourse
+ENV CONTAINER_APP_UID 200
+ENV CONTAINER_APP_GROUP discourse
+ENV CONTAINER_APP_GID 200
```
gregory-schiano, line 11: Why do you remove that in a PR which is not related to that at all ?
It was probably added for a purpose, I think by Niels. Did you discuss that with him ?

Globally speaking I see too many PRs which are changing things out of the scope of the ticket, leading to some changes merged under the wood.

tests/unit/test_charm.py
jdkandersson: I think you are referring to the PR that added the add-admin or similar action. The difference on that one was there was an integration test and the conversation was about to degree to which the final state should be checked, in this case there wasn't an integration test so the cases are different.

Also, anonymizing user data I assume is in response to a user request that is covered by regulation. Perhaps that warrants some thinking about business criticality compared to an action that adds an admin user which is probably not covered by regulation as much?

PR #40, opened 2023-03-07
Upgrade to discourse 2.8.14

arturo-seijas:


src/charm.py
```diff
@@ -389,26 +383,40 @@ def _config_changed(self, event: HookEvent) -> None:
...
-                [script],
-                environment=self._create_discourse_environment_settings(),
-                working_dir="/srv/discourse/app",
-            )
+        if self._is_config_valid() and self.model.unit.is_leader() and not current_plan.services:
+            env_settings = self._create_discourse_environment_settings()
             try:
-                stdout, _ = process.wait_output()
-                logger.debug("%s stdout: %s", script, stdout)
+                self.model.unit.status = MaintenanceStatus("Running migrations")
```
jdkandersson, line 392: Should we include an upgrade test for this?

yanksyoon (approved):
LGTM :)

src/charm.py
arturo-seijas: I closed this PR for now but I'm replying anyway.
The migrations are already run on installation. The was Rails handles migrations make unnecessary to run an upgrade test. All changes are applied iteratively in a deterministic order (the migration file names are prefixed by a timestamp) both for installation and upgrade. In the second case, rails will remember the last applied change and handle the rest.

PR #43, opened 2023-03-09
Upgrade to discourse 2.8.14

### LINKED ISSUE #30: Discourse: Fix the issues found by Trivy scanning
### Bug Description

By default, the docker images will be scanned by Trivy in the integration test and will fail if some critical or high-severity vulnerabilities are found.

While scanning the discourse-image, there were found vulnerabilities.

To not impact the workflow, the Trivy configuration was set to not fail even finding those issues.

They should be analyzed and fixed, if possible, or ignored.

Acceptance criteria:

Fixed issues.

Trivy configuration set to ignore unfixed, if needed.

### To Reproduce

Any PR triggers integration tests that include a Trivy scan that showcases the behavior.

### Environment

CI

### Relevant log output

```shell
n/a
```

### Additional context

_No response_

arturo-seijas:
Closes #30


gtrkiller (dismissed):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   integration-tests:
-    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@debug-2
```
jdkandersson, line 8: I assume we're not closing this until this points back at `main`?

src/charm.py
```diff
@@ -392,21 +390,36 @@ def _config_changed(self, event: HookEvent) -> None:
...
+                    self.model.unit.status = MaintenanceStatus("Compiling assets")
+                    script = f"{SCRIPT_PATH}/pod_setup.sh"
+                    process = container.exec(
+                        [script],
+                        environment=env_settings,
+                        working_dir="/srv/discourse/app",
+                        user="discourse",
+                    )
+                    process.wait_output()
+                    self.model.unit.status = MaintenanceStatus("Compiling assets")
```
jdkandersson, line 406: This status already seems to be set on line 397?

src/charm.py
```diff
@@ -392,21 +390,36 @@ def _config_changed(self, event: HookEvent) -> None:
...
-                self.model.unit.status = MaintenanceStatus("Compiling assets")
-                script = f"{SCRIPT_PATH}/pod_setup.sh"
-                process = container.exec(
-                    [script],
-                    environment=self._create_discourse_environment_settings(),
-                    working_dir="/srv/discourse/app",
-                    user="discourse",
-                )
-                process.wait_output()
+                if not current_plan.services:
```
jdkandersson, line 396: This looks like the nesting is getting quite deep, should this be in a separate function?

.github/workflows/integration_test.yaml
arturo-seijas: Yes

src/charm.py
arturo-seijas: Fixed

src/charm.py
arturo-seijas: I'll be doing some refactoring as part of this PR. I'll close the comment for now

src/charm.py
```diff
@@ -327,15 +327,16 @@ def _should_run_s3_migration(self, current_plan: Dict, s3info: Optional[S3Info])
...
-            # Or S3 is enabled and one S3 parameter has changed
-            self.config.get("s3_enabled")
-            and s3info
-            and (
-                s3info.enabled != self.config.get("s3_enabled")
-                or s3info.region != self.config.get("s3_region")
-                or s3info.bucket != self.config.get("s3_bucket")
-                or s3info.endpoint != self.config.get("s3_endpoint")
+        return self.config.get("s3_enabled") and (
+            not current_plan.services  # type: ignore
```
jdkandersson, line 331: @benhoyt any thoughts on this? @arturo-seijas I'm guessing the comment on line 329 explains the reason for the type: ignore here?

src/charm.py
```diff
@@ -430,7 +443,7 @@ def _config_changed(self, event: HookEvent) -> None:
             container.add_layer(SERVICE_NAME, layer_config, combine=True)
             container.pebble.replan_services()
             self.ingress.update_config(self._make_ingress_config())
-            self._config_force_https()
+            # self._config_force_https()
```
jdkandersson, line 446: Can this line be removed rather than commented out? Generally it would be better not to have commented out code as it makes maintenance more difficult

src/charm.py
```diff
@@ -458,17 +471,18 @@ def _on_database_changed(self, event: pgsql.DatabaseChangedEvent) -> None:  # ty
...
-        self._stored.db_user = None
-        self._stored.db_password = None
-        self._stored.db_host = None
-        if event.master is not None:
+        if event.master is None:
+            self._stored.db_name = None
+            self._stored.db_user = None
+            self._stored.db_password = None
+            self._stored.db_host = None
+        elif event.master is not None:
```
jdkandersson, line 479: Could this just be an `else`?

src/charm.py
```diff
@@ -458,17 +471,18 @@ def _on_database_changed(self, event: pgsql.DatabaseChangedEvent) -> None:  # ty
...
             self._stored.db_name = event.master.dbname
             self._stored.db_user = event.master.user
             self._stored.db_password = event.master.password
             self._stored.db_host = event.master.host
 
-        self._config_changed(event)
+        # self._config_changed(event)
```
jdkandersson, line 485: Similar comment re commented out code

tests/integration/conftest.py
```diff
@@ -138,7 +138,7 @@ async def app_fixture(
         model.add_relation(app_name, "redis-k8s"),
         model.add_relation(app_name, "nginx-ingress-integrator"),
     )
-    await model.wait_for_idle(status="active")
+    await model.wait_for_idle(status="active", raise_on_error=False)
```
jdkandersson, line 141: Is this also due to postgres-k8s stability? @yanksyoon and @delgod FYI here as well

arturo-seijas: Yes

src/charm.py
arturo-seijas: I was just testing some scenarios

src/charm.py
benhoyt: No, I don't think we should be `type: ignore`ing here. The type signature of `_should_run_s3_migration` is incorrect: `current_plan` should be a `pebble.Plan`, not a `Dict`. That gets rid of one mypy error. The other error you can fix by converting the result to a proper bool instead of an `and`ed truthy value, using `bool(result)`.

I've removed the `type: ignore` and made the following changes and it type checks nicely:

```diff
diff --git a/src/charm.py b/src/charm.py
index dec3ad0..c82417f 100755
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,7 +18,7 @@ from ops.charm import ActionEvent, CharmBase, HookEvent
 from ops.framework import StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
-from ops.pebble import ExecError
+from ops.pebble import ExecError, Plan
 
 logger = logging.getLogger(__name__)
 pgsql = ops.lib.use("pgsql", 1, "postgresql-charmers@lists.launchpad.net")
@@ -318,7 +318,7 @@ class DiscourseCharm(CharmBase):
         }
         return layer_config
 
-    def _should_run_s3_migration(self, current_plan: Dict, s3info: Optional[S3Info]) -> bool:
+    def _should_run_s3_migration(self, current_plan: Plan, s3info: Optional[S3Info]) -> bool:
         """Determine if the S3 migration is to be run.
 
         Args:
@@ -328,9 +328,8 @@ class DiscourseCharm(CharmBase):
         Returns:
             If no services are planned yet (first run) or S3 settings have changed.
         """
-        # Properly type checks would require defining a complex TypedMap for the pebble plan
-        return self.config.get("s3_enabled") and (
-            not current_plan.services  # type: ignore
+        result = self.config.get("s3_enabled") and (
+            not current_plan.services
             or (
                 s3info
                 and (
@@ -341,6 +340,7 @@ class DiscourseCharm(CharmBase):
                 )
             )
         )
+        return bool(result)
 
     def _are_db_relations_ready(self) -> bool:
         """Check if the needed database relations are established.
```

jdkandersson: Excellent, thank you @benhoyt!

yanksyoon (approved):
LGTM!

discourse.Dockerfile
```diff
@@ -46,16 +45,11 @@ RUN ln -s /usr/share/zoneinfo/UTC /etc/localtime \
...
-# vital to trigger js assets compression using node. Manually create the
-# terser command if it does not exist. Please remove this line if the
-# base image is >= 22.04. Also, please consider removing the node-uglify
-# when upgrading to discourse >= 2.8.0, since the node-uglify is not
-# required to trigger node js assets compression, only terser will do fine.
-    && which terser || ln -s $(which uglifyjs.terser) /usr/local/bin/terser \
-# Run build process.
+    && curl --silent --location https://deb.nodesource.com/setup_18.x | bash - \
+    && apt-get update \
+    && apt-get install -y nodejs \
```
gregory-schiano, line 50: Why not installing nodejs with the packages ?

src/charm.py
```diff
@@ -357,21 +358,51 @@ def _are_db_relations_ready(self) -> bool:
...
+                    user="discourse",
+                )
+                process.wait_output()
+            self.model.unit.status = MaintenanceStatus("Compiling assets")
+            process = container.exec(
+                [f"{DISCOURSE_PATH}/bin/bundle", "exec", "rake", "assets:precompile"],
+                environment=env_settings,
+                working_dir=DISCOURSE_PATH,
+                user="discourse",
+            )
```
gregory-schiano, line 388: Is the install/upgrade even triggered if the container is rescheduled ?
A reschedule will "lose" the precompiled asset, if the event is not re-triggered we'll get an issue with this change

gregory-schiano: Also.. shouldn't all units run the precompile assets ? I don't think they're in a folder shared by all units, therefore each unit should have it local precompiled assets no ?

tests/integration/test_charm.py
```diff
@@ -155,7 +155,7 @@ async def test_setup_discourse(
     logger.info("Getting account activation page")
 
     response = session.get(
-        f"{discourse_url}/u/activate-account/{email_token}",
+        f"{discourse_url}/u/password-reset//{email_token}",
```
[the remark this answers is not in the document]
arturo-seijas: I can't see that line anymore. Perhaps you are looking at a previous revision?

discourse.Dockerfile
arturo-seijas: Which packages are you referring to?

src/charm.py
arturo-seijas: Exactly. That's the current behaviour, but the migrations are only run in the leader unit

gregory-schiano: Oh yes indeed.
The comment is still valid concerning the "install/upgrade" event even if it seems that the upgrade event is triggered when the container is rescheduled

tests/integration/test_charm.py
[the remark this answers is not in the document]
gregory-schiano: Well indeed... don't know what github did there when I opened the tab.
Closed sorry for the noise

discourse.Dockerfile
arturo-seijas: The node version in the default package list doesn't match the one we need. Additionally, I need curl, which is not installed by default.

gregory-schiano: Ok got it
1) Install curl
2) Add node repo
3) Install node

I didn't realise the `curl | bash` before was adding a new deb repo

gregory-schiano (approved):
LGTM
As discussed, I approve this PR as is, and depending on the feedback from juju on the expected events when container is reschedule, we'll adapt in a different PR if needed

PR #44, opened 2023-03-17
fix: Discourse category create internal server error fix

yanksyoon:
This PR fixes the high priority issue that is caused by Discourse application bug by upgrading the discourse patch version.


discourse.Dockerfile
```diff
@@ -1,10 +1,10 @@
 FROM ubuntu:focal
 
 # Used in Launchpad OCI Recipe build to tag the image.
-LABEL org.label-schema.version=v2.7.10
+LABEL org.label-schema.version=v2.7.13
```
yanksyoon, line 4: latest version in 2.7 from github. https://github.com/discourse/discourse/tags?after=v2.8.4

src/charm.py
```diff
@@ -34,18 +34,18 @@
 }
 THROTTLE_LEVELS["permissive"] = {
     "DISCOURSE_MAX_REQS_PER_IP_MODE": "warn+block",
-    "DISCOURSE_MAX_REQS_PER_IP_PER_MINUTE": 1000,
-    "DISCOURSE_MAX_REQS_PER_IP_PER_10_SECONDS": 100,
-    "DISCOURSE_MAX_USER_API_REQS_PER_MINUTE": 400,
-    "DISCOURSE_MAX_ASSET_REQS_PER_IP_PER_10_SECONDS": 400,
+    "DISCOURSE_MAX_REQS_PER_IP_PER_MINUTE": "1000",
```
yanksyoon, line 37: pebble accepts environment variables as map[str, str]

discourse.Dockerfile
arturo-seijas: Perhaps it's worth testing if https://github.com/canonical/discourse-k8s-operator/pull/43 does still crash and upgrade to the last version 2 that fixes all the existing critital and high vulnerabilities detected by trivy

yanksyoon: Sure, let me move the tests for creating category to #43. Would this be close to merging? since this is high priority fix, I was hoping to get it merged within today :)

arturo-seijas: I've just pushed some changes. I suggest we see if the integration tests pass before doing anything.

yanksyoon: Sure, i'll create another branch and create a PR towards #43 :)

yanksyoon:
Merged #45 into #43 which does the same thing :)

PR #45, opened 2023-03-17
test: add create category test

yanksyoon:
This PR adds tests to check the patched status of create category internal server error from discourse application.
Note that this is to be merged not to main but to `upgrade-to-discourse-2.8.14` branch, for #43.


arturo-seijas (approved):
LGTM

PR #47, opened 2023-03-20
fix: restart after assets compile

yanksyoon:
This PR addresses the problem where the service is not restarted after compiling assets, resulting in serving non-compiled assets, that will result in 404 while loading assets.

`container.replan()` fails to actually stop and start the service, hence the `container.stop()` followed by `container.start()`.

```
  File "./src/charm.py", line 451, in _reload_configuration
    container.replan()
  File "/var/lib/juju/agents/unit-discourse-k8s-0/charm/venv/ops/model.py", line 1877, in replan
    self._pebble.replan_services()
  File "/var/lib/juju/agents/unit-discourse-k8s-0/charm/venv/ops/pebble.py", line 1571, in replan_services
    return self._services_action('replan', [], timeout, delay)
  File "/var/lib/juju/agents/unit-discourse-k8s-0/charm/venv/ops/pebble.py", line 1652, in _services_action
    raise ChangeError(change.err, change)
ops.pebble.ChangeError: cannot perform the following tasks:
- Start service "discourse" (cannot start service while terminating)
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,7 +18,7 @@ from ops.charm import ActionEvent, CharmBase, HookEvent
 from ops.framework import StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
-from ops.pebble import ExecError, Plan
+from ops.pebble import ExecError, ExecProcess, Plan
 
 logger = logging.getLogger(__name__)
 pgsql = ops.lib.use("pgsql", 1, "postgresql-charmers@lists.launchpad.net")
@@ -372,7 +372,7 @@ class DiscourseCharm(CharmBase):
         try:
             if self.model.unit.is_leader():
                 self.model.unit.status = MaintenanceStatus("Executing migrations")
-                process = container.exec(
+                process: ExecProcess = container.exec(
                     [f"{DISCOURSE_PATH}/bin/bundle", "exec", "rake", "--trace", "db:migrate"],
                     environment=env_settings,
                     working_dir=DISCOURSE_PATH,
@@ -380,7 +380,7 @@ class DiscourseCharm(CharmBase):
                 )
                 process.wait_output()
             self.model.unit.status = MaintenanceStatus("Compiling assets")
-            process = container.exec(
+            process: ExecProcess = container.exec(
                 [f"{DISCOURSE_PATH}/bin/bundle", "exec", "rake", "assets:precompile"],
                 environment=env_settings,
                 working_dir=DISCOURSE_PATH,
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -446,7 +446,8 @@ def _reload_configuration(self) -> None:
         if self._is_config_valid() and container.can_connect():
             layer_config = self._create_layer_config()
             container.add_layer(SERVICE_NAME, layer_config, combine=True)
-            container.pebble.replan_services()
+            container.stop(SERVICE_NAME)
```
arturo-seijas, line 449: Can you add a comment explaining the change and a link to the opened issue, if any?

yanksyoon: Done :)

arturo-seijas (dismissed):
LGTM

arturo-seijas (approved):
LGTM

src/charm.py
```diff
@@ -446,7 +446,11 @@ def _reload_configuration(self) -> None:
         if self._is_config_valid() and container.can_connect():
             layer_config = self._create_layer_config()
             container.add_layer(SERVICE_NAME, layer_config, combine=True)
-            container.pebble.replan_services()
+            # Currently, pebble replan will cause
+            # ChangeError("cannot start service while terminating".)
+            # Link to issue: https://github.com/canonical/pebble/issues/206
+            container.stop(SERVICE_NAME)
+            container.start(SERVICE_NAME)
```
mthaddon, line 453: This seems like we're going to be doing a stop and a start any time `_reload_configuration` is called, which might happen on any `config-changed`. I wonder if it's possible to only do a stop/start if we have recompiled the assets? Or do we think a stop/start is definitely needed any time `config-changed` is called?

gtrkiller (approved):
LGTM

src/charm.py
yanksyoon: That is a good point! Would limiting the scope of this PR to the simple fix for the moment and creating the refactor as another issue work for you? Since container.replan() also internally stop/starts the services the effect is currently the same. From my limited knowledge, I think it may require a bit more planning to separate out the flows in which the stop/start service is definitely required.
What do you think? :)

mthaddon: Yes, I think it'd be reasonable to do this in a separate PR as long as we get to it soon (before we begin production rollouts of this charm.

It might not be too complicated, I think. As I understand it pebble will restart a service if any environment variable you're passing in has changed (this will need testing). We could simply create a `DISCOURSE_ASSETS_LAST_UPDATED` environment variable with a string rendering of timestamp, for example, and update that any time we want the service to be restarted, and then continue with `container.replan_services`.

Anyway, feel free to close out this conversation and follow up in a separate PR.

jdkandersson (approved):

PR #48, opened 2023-03-20
Update tutorial to include adding an admin user

arturo-seijas:
Update the getting started tutorial to include adding an admin user.


yanksyoon (approved):
Would it be worth duplicating the same information in how-to-guides as well? Quite trivial and it's defined in `actions.yaml` as well so I don't mind it not being there either :)
LGTM!

gregory-schiano (approved):
LGTM

PR #49, opened 2023-03-22
Wait until the discourse setup is complete before starting the discourse server

weiiwang01:
In the `_on_database_changed` callback, the function actually calls  `self._reload_configuration`, which starts the discourse server. However, since `_set_up_discourse` is deferred until the database is established, this causes `_reload_configuration` to run before `_set_up_discourse`, resulting in the discourse server starting without the database table being created.

Add a check in `_on_database_changed` to wait until discourse setup is finished.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,7 +5,6 @@
 """Charm for Discourse on kubernetes."""
 import logging
 import os.path
-import pathlib
 from collections import defaultdict, namedtuple
 from typing import Any, Dict, List, Optional
 
@@ -59,7 +58,7 @@ REQUIRED_S3_SETTINGS = ["s3_access_key_id", "s3_bucket", "s3_region", "s3_secret
 SCRIPT_PATH = "/srv/scripts"
 SERVICE_NAME = "discourse"
 SERVICE_PORT = 3000
-SETUP_COMPLETED_FLAG_FILE = pathlib.Path("/run/discourse-k8s-operator/setup_completed")
+SETUP_COMPLETED_FLAG_FILE = "/run/discourse-k8s-operator/setup_completed"
 
 
 class DiscourseCharm(CharmBase):
@@ -134,12 +133,13 @@ class DiscourseCharm(CharmBase):
         Returns:
             True if the _set_up_discourse process has finished.
         """
-        return SETUP_COMPLETED_FLAG_FILE.is_file()
+        container = self.unit.get_container(SERVICE_NAME)
+        return container.can_connect() and container.exists(SETUP_COMPLETED_FLAG_FILE)
 
     def _set_setup_completed(self) -> None:
         """Mark the _set_up_discourse process as completed."""
-        SETUP_COMPLETED_FLAG_FILE.parent.mkdir(exist_ok=True)
-        SETUP_COMPLETED_FLAG_FILE.touch(exist_ok=True)
+        container = self.unit.get_container(SERVICE_NAME)
+        container.push(SETUP_COMPLETED_FLAG_FILE, "", make_dirs=True)
 
     def _is_config_valid(self) -> bool:
         """Check that the provided config is valid.
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,6 +5,7 @@
 """Charm for Discourse on kubernetes."""
 import logging
 import os.path
+import pathlib
 from collections import defaultdict, namedtuple
 from typing import Any, Dict, List, Optional
 
@@ -58,6 +59,7 @@ REQUIRED_S3_SETTINGS = ["s3_access_key_id", "s3_bucket", "s3_region", "s3_secret
 SCRIPT_PATH = "/srv/scripts"
 SERVICE_NAME = "discourse"
 SERVICE_PORT = 3000
+SETUP_COMPLETED_FLAG_FILE = pathlib.Path("/run/discourse-k8s-operator/setup_completed")
 
 
 class DiscourseCharm(CharmBase):
@@ -75,7 +77,6 @@ class DiscourseCharm(CharmBase):
             db_user=None,
             db_password=None,
             db_host=None,
-            setup_completed=False,
             redis_relation={},
         )
         self.ingress = IngressRequires(self, self._make_ingress_config())
@@ -126,6 +127,19 @@ class DiscourseCharm(CharmBase):
         return (
             self.config["external_hostname"] if self.config["external_hostname"] else self.app.name
         )
+
+    def _is_setup_completed(self) -> bool:
+        """Check if the _set_up_discourse process has finished.
+
+        Returns:
+            True if the _set_up_discourse process has finished.
+        """
+        return SETUP_COMPLETED_FLAG_FILE.is_file()
+
+    def _set_setup_completed(self) -> None:
+        """Mark the _set_up_discourse process as completed."""
+        SETUP_COMPLETED_FLAG_FILE.parent.mkdir(exist_ok=True)
+        pathlib.Path("/run/discourse-k8s-operator/setup_completed").touch(exist_ok=True)
 
     def _is_config_valid(self) -> bool:
         """Check that the provided config is valid.
@@ -388,7 +402,7 @@ class DiscourseCharm(CharmBase):
                 user="discourse",
             )
             precompile_process.wait_output()
-            self._stored.setup_completed = True
+            self._set_setup_completed()
             get_version_process = container.exec(
                 [f"{DISCOURSE_PATH}/bin/rails", "runner", "puts Discourse::VERSION::STRING"],
                 environment=env_settings,
@@ -453,7 +467,7 @@ class DiscourseCharm(CharmBase):
 
     def _reload_configuration(self) -> None:
         # mypy has some trouble with dynamic attributes
-        if not self._stored.setup_completed:  # type: ignore
+        if not self._is_setup_completed():
             logger.info("Defer starting the discourse server until discourse setup completed")
             return
         container = self.unit.get_container(SERVICE_NAME)
--- a/src/charm.py
+++ b/src/charm.py
@@ -75,6 +75,7 @@ class DiscourseCharm(CharmBase):
             db_user=None,
             db_password=None,
             db_host=None,
+            setup_completed=False,
             redis_relation={},
         )
         self.ingress = IngressRequires(self, self._make_ingress_config())
@@ -387,6 +388,7 @@ class DiscourseCharm(CharmBase):
                 user="discourse",
             )
             precompile_process.wait_output()
+            self._stored.setup_completed = True
         except ExecError as cmd_err:
             logger.exception("Setting up discourse failed with code %d.", cmd_err.exit_code)
             raise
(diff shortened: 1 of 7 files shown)


arturo-seijas (dismissed):
LGTM

arturo-seijas (commented):
What happens is the pod is restarted? I know the assets will be compiled again but, will the state be restored?

weiiwang01:
> restarted

The `upgrade-charm` event will trigger the `_set_up_discourse` to set the `setup_completed` flag.

weiiwang01:
> What happens is the pod is restarted? I know the assets will be compiled again but, will the state be restored?

The `upgrade-charm` event will trigger the `_set_up_discourse` to set the `setup_completed` flag.

arturo-seijas:
> > What happens is the pod is restarted? I know the assets will be compiled again but, will the state be restored?
> 
> The `upgrade-charm` event will trigger the `_set_up_discourse` to set the `setup_completed` flag.

I mean, in the event of a restart when Discourse is running properly. Will `setup_completed` keep its value `true`?

weiiwang01:
> > > What happens is the pod is restarted? I know the assets will be compiled again but, will the state be restored?
> > 
> > 
> > The `upgrade-charm` event will trigger the `_set_up_discourse` to set the `setup_completed` flag.
> 
> I mean, in the event of a restart when Discourse is running properly. Will `setup_completed` keep its value `true`?

When the container is rescheduled, the discourse is terminated too. At the begining of the new container, `setup_completed` will be `False`. But shortly, the `_set_up_discourse` function will start and set the `setup_completed` to `True`. After that, discourse server will start.

arturo-seijas (dismissed):
LGTM

gregory-schiano (dismissed):
I approve because there's a previous usage of StoredState. But since storedstate is deprecated it'll require rework.
We should also always keep in mind that storedstate is reset when the pod is rescheduled (unless under some deployment conditions).

The proper way to fix this will be to use the peer databag

src/charm.py
```diff
@@ -75,6 +75,7 @@ def __init__(self, *args):
             db_user=None,
```
jdkandersson, line 75: Do you think an integration test that demonstrates this issue is worth it? Otherwise someone might just remove the code in the future again and there wouldn't be any test failure?

jdkandersson: Discussed with Weii, should be ok without an integration test since the user can't actually tell this problem is happening

weiiwang01:
> I approve because there's a previous usage of StoredState. But since storedstate is deprecated it'll require rework. We should also always keep in mind that storedstate is reset when the pod is rescheduled (unless under some deployment conditions).
> 
> The proper way to fix this will be to use the peer databag

I have some mixed feeling about the StoredState, the StoredState is deprecated mostly because the state is not persisted during container rescheduling. However sometime, it's handy to save some information inside the container and that's not required to be persisted during container rescheduling. For example, the `setup_completed` in this PR can be reset during container rescheduling without any problem.

gregory-schiano:
> > I approve because there's a previous usage of StoredState. But since storedstate is deprecated it'll require rework. We should also always keep in mind that storedstate is reset when the pod is rescheduled (unless under some deployment conditions).
> > The proper way to fix this will be to use the peer databag
> 
> I have some mixed feeling about the StoredState, the StoredState is deprecated mostly because the state is not persisted during container rescheduling. However sometime, it's handy to save some information inside the container and that's not required to be persisted during container rescheduling. For example, the `setup_completed` in this PR can be reset during container rescheduling without any problem.

In such cases you touch a file in the container (and not in a mounted volume of the container) and check its presence, this won't persist after rescheduling
You can also reset your flag store in peer databag during the install hook

src/charm.py
```diff
@@ -442,6 +444,10 @@ def _config_changed(self, event: HookEvent) -> None:
             self.model.unit.status = ActiveStatus()
 
     def _reload_configuration(self) -> None:
+        # mypy has some trouble with dynamic attributes
+        if not self._stored.setup_completed:  # type: ignore
+            logger.info("Defer starting the discourse server until discourse setup completed")
+            return
```
yanksyoon, line 450: Out of curiosity, is the event truly deferred? My understanding is that event.defer() needs to be called for it to be "re-queued". If it is not deferred, what triggers the supposedly deferred event to run again?

yanksyoon (dismissed):
LGTM! :)

src/charm.py
weiiwang01: The event is not deferred, but the `_reload_configuration` function will run in multiple hooks for example `config-changed`, `discourse-pebble-ready`, `database-relateion-change`. One of them will start the discourse server eventually.

arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM

gregory-schiano (dismissed):
LGTM

mthaddon:
> > > > What happens is the pod is restarted? I know the assets will be compiled again but, will the state be restored?
> > > 
> > > 
> > > The `upgrade-charm` event will trigger the `_set_up_discourse` to set the `setup_completed` flag.
> > 
> > 
> > I mean, in the event of a restart when Discourse is running properly. Will `setup_completed` keep its value `true`?
> 
> When the container is rescheduled, the discourse is terminated too. At the begining of the new container, `setup_completed` will be `False`. But shortly, the `_set_up_discourse` function will start and set the `setup_completed` to `True`. After that, discourse server will start.

I'm not sure I follow how `setup_completed` will be `False` at the beginning of the new container. We're using stored state for this and since we're setting `use_juju_for_storage=True` that will be stored in the juju controller. So before the container is rescheduled this value would be `True` and when the container is rescheduled I don't see anything here that would set it to `False`. Am I missing something?

gregory-schiano:
> > > > > What happens is the pod is restarted? I know the assets will be compiled again but, will the state be restored?
> > > > 
> > > > 
> > > > The `upgrade-charm` event will trigger the `_set_up_discourse` to set the `setup_completed` flag.
> > > 
> > > 
> > > I mean, in the event of a restart when Discourse is running properly. Will `setup_completed` keep its value `true`?
> > 
> > 
> > When the container is rescheduled, the discourse is terminated too. At the begining of the new container, `setup_completed` will be `False`. But shortly, the `_set_up_discourse` function will start and set the `setup_completed` to `True`. After that, discourse server will start.
> 
> I'm not sure I follow how `setup_completed` will be `False` at the beginning of the new container. We're using stored state for this and since we're setting `use_juju_for_storage=True` that will be stored in the juju controller. So before the container is rescheduled this value would be `True` and when the container is rescheduled I don't see anything here that would set it to `False`. Am I missing something?

Ow ! I didn't see we were setting it in the main ! That's tricky !

mthaddon:
> > > > > > What happens is the pod is restarted? I know the assets will be compiled again but, will the state be restored?
> > > > > 
> > > > > 
> > > > > The `upgrade-charm` event will trigger the `_set_up_discourse` to set the `setup_completed` flag.
> > > > 
> > > > 
> > > > I mean, in the event of a restart when Discourse is running properly. Will `setup_completed` keep its value `true`?
> > > 
> > > 
> > > When the container is rescheduled, the discourse is terminated too. At the begining of the new container, `setup_completed` will be `False`. But shortly, the `_set_up_discourse` function will start and set the `setup_completed` to `True`. After that, discourse server will start.
> > 
> > 
> > I'm not sure I follow how `setup_completed` will be `False` at the beginning of the new container. We're using stored state for this and since we're setting `use_juju_for_storage=True` that will be stored in the juju controller. So before the container is rescheduled this value would be `True` and when the container is rescheduled I don't see anything here that would set it to `False`. Am I missing something?
> 
> Ow ! I didn't see we were setting it in the main ! That's tricky !

Yeah, you basically have to do that otherwise your state doesn't persist across container reschedule events. Obviously in this particular case that would have worked for us, but in other cases that we're using it for that would be a problem :(

src/charm.py
```diff
@@ -127,6 +128,19 @@ def _get_external_hostname(self) -> str:
...
+
+        Returns:
+            True if the _set_up_discourse process has finished.
+        """
+        return SETUP_COMPLETED_FLAG_FILE.is_file()
+
+    def _set_setup_completed(self) -> None:
+        """Mark the _set_up_discourse process as completed."""
+        SETUP_COMPLETED_FLAG_FILE.parent.mkdir(exist_ok=True)
+        pathlib.Path("/run/discourse-k8s-operator/setup_completed").touch(exist_ok=True)
```
jdkandersson, line 142: Should this use the constant from the top of the file?

weiiwang01: Yes, updated, thanks!

src/charm.py
```diff
@@ -127,6 +128,19 @@ def _get_external_hostname(self) -> str:
             self.config["external_hostname"] if self.config["external_hostname"] else self.app.name
         )
 
+    def _is_setup_completed(self) -> bool:
```
arturo-seijas, line 131: It is my understanding that this will create the file in the `charm`container. I think it'd be better to leverage pebble to push and pull a file to/from the discourse container, to guarantee a container restart triggers the generation of static assets.

weiiwang01: Good point, I updated the code to put the file in discourse container instead.

arturo-seijas (dismissed):
LGTM

gregory-schiano (dismissed):
LGTM

mthaddon:
Is it worth removing the `SETUP_COMPLETED_FLAG_FILE` file while compiling assets? My understanding is the assets may get recompiled if config changes, and it would be good to reflect that they're not compiled until that's been done in case something else is checking for this.

weiiwang01:
> Is it worth removing the `SETUP_COMPLETED_FLAG_FILE` file while compiling assets? My understanding is the assets may get recompiled if config changes, and it would be good to reflect that they're not compiled until that's been done in case something else is checking for this.

The `_set_up_discourse` function is only triggered by `install` and `upgrade-charm` event, I don't think the asset will be recompiled during a config change.

jdkandersson (commented):
I'm ok with the changes, looks like there is still some questions to answer so I'll hold off on approval for now

tests/unit/test_charm.py
```diff
@@ -204,29 +219,30 @@ def test_config_changed_when_s3_and_no_bucket_invalid(self):
...
+            self.harness.disable_hooks()
+            self.harness.set_leader(True)
+            self._add_database_relations()
+            self.harness.update_config(
+                {
+                    "s3_access_key_id": "3|33+",
+                    "s3_bucket": "who-s-a-good-bucket?",
+                    "s3_enabled": True,
+                    "s3_endpoint": "s3.endpoint",
+                    "s3_region": "the-infinite-and-beyond",
```
yanksyoon, line 240: 😂

yanksyoon (dismissed):
LGTM!

tests/unit/test_charm.py
amandahla: Nice 😂

amandahla (dismissed):
LGTM

arturo-seijas (approved):
LGTM

jdkandersson (approved):

PR #63, opened 2023-05-22
Docs/postgresql channel

merkata:
Deploying `postgresql-k8s` now fails with
```
merkata@mastodon:~$ juju deploy postgresql-k8s
ERROR resolving retry error: No revision was found in the Store.
```
If you want the version pre-14 (so 14/stable as described in the [doc](https://charmhub.io/postgresql-k8s/docs/t-deploy-postgresql)), you'd need to pass in `latest/stable`. _Note that we're not using 14/stable now as it doesn't support extensions and the psql charm blocks._


nrobinaubertin (approved):

yanksyoon (approved):
LGTM!

PR #67, opened 2023-05-25
docs(tutorial): Use valid param length

### LINKED ISSUE #66: Failing add-admin action due to short password in the docs
### Bug Description

The docs specify to use `somepwd` when setting up an admin account, however that errors out, as the password requirement is to be more than 10 characters. The doc needs to be updated with a password suggestion that works.

### To Reproduce

1. Follow the guide.
2. Fail on the admin action by following the guide.

### Environment

juju+microk8s

### Relevant log output

```shell
Password is too short (minimum is 10 characters)
```

### Additional context

_No response_

merkata:
closes #66 - when you follow the guide ans input `somepwd` as a password for the admin action, it fails, as the password checks that the length is a 10 characters minimum

### DIFF
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -81,4 +81,4 @@ Discourse will be deployed with `discourse-k8s` as default hostname. In order to
 
 After that, visit `http://discourse-k8s` to reach Discourse.
 
-This charm allows you to add a admin user by executing the corresponding action. Instead of interacting with the Discourse UI just run `juju run-action discourse-k8s/0 add-admin-user email=email@example.com password=somepwd --wait` and it will be registered and validated.
+This charm allows you to add a admin user by executing the corresponding action. Instead of interacting with the Discourse UI just run `juju run-action discourse-k8s/0 add-admin-user email=email@example.com password=somepassword --wait` and it will be registered and validated.
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -81,4 +81,4 @@ Discourse will be deployed with `discourse-k8s` as default hostname. In order to
 
 After that, visit `http://discourse-k8s` to reach Discourse.
 
-This charm allows you to add a admin user by executing the corresponding action. Instead of interacting with the Discourse UI just run `juju run-action discourse-k8s/0 add-admin-user email=email@example.com password=<secure password> --wait` and it will be registered and validated. Note that the password needs to be at minimum 10 characters long, otherwise the action will fail.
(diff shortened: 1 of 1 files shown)


docs/tutorials/getting-started.md
```diff
@@ -81,4 +81,4 @@ Discourse will be deployed with `discourse-k8s` as default hostname. In order to
 
 After that, visit `http://discourse-k8s` to reach Discourse.
 
-This charm allows you to add a admin user by executing the corresponding action. Instead of interacting with the Discourse UI just run `juju run-action discourse-k8s/0 add-admin-user email=email@example.com password=somepwd --wait` and it will be registered and validated.
+This charm allows you to add a admin user by executing the corresponding action. Instead of interacting with the Discourse UI just run `juju run-action discourse-k8s/0 add-admin-user email=email@example.com password=somepassword --wait` and it will be registered and validated.
```
jdkandersson, line 84: should this be `<secure password>` to indicate that it should be replaced by a secure password?

merkata: Good point, could be - I was replacing `somepwd` with `somepassword` as following the guide produces an error (somepasswd < 10 characters and the action fails). So `somepassword` works as an example. I could place `<secure password>`, though I need to also point out that it has to also be more than 10 characters. What do you think?

jdkandersson: Yes that is a good approach. It is important not to encourage setting insecure passwords in our tutorials

gregory-schiano:
Could it be interesting to provide the command to check the output of the action if it fails ? I'd expect an explicit error message when the password length is to short

merkata:
> Could it be interesting to provide the command to check the output of the action if it fails ? I'd expect an explicit error message when the password length is to short

If I understand correctly - the command is the one described, just the password supplied is less than 10 characters, then you get a lengthy output by juju telling you the command failed. I could put the error log here if it adds value, still I believe that if someone uses the action according to the tutorial (it says a password needs to be at least 10 characters), they should be fine. I want to avoid too much text that covers the corner case of someone not reading the tutorial in full (i.e. not reading or skipping the password requirement).

gregory-schiano:
> > Could it be interesting to provide the command to check the output of the action if it fails ? I'd expect an explicit error message when the password length is to short
> 
> If I understand correctly - the command is the one described, just the password supplied is less than 10 characters, then you get a lengthy output by juju telling you the command failed. I could put the error log here if it adds value, still I believe that if someone uses the action according to the tutorial (it says a password needs to be at least 10 characters), they should be fine. I want to avoid too much text that covers the corner case of someone not reading the tutorial in full (i.e. not reading or skipping the password requirement).

Then I suggest (if possible) to highlight a bit more the 10 chars rule, to ensure this is clearly visible for the users

jdkandersson (dismissed):

nrobinaubertin (approved):

arturo-seijas (approved):

yanksyoon (approved):
LGTM!

PR #75, opened 2023-06-26
Create a rockfile to replace the current Dockerfile

nrobinaubertin:
### Overview

The goal of this PR is to replace the current dockerfile with a rockfile.

Notes:
- I decided to *not* use a `bare` base, opting for simplicity for now. In the same spirit, I'm using the root user at runtime.  
- The integration tests rely on [this draft PR](https://github.com/canonical/operator-workflows/pull/176) so that means that it's not really mergeable until the latter is.  
- Integration tests rely on `raise_on_error=False` sometimes, I want to work on this in a further PR. 

### Checklist


tests/integration/conftest.py
```diff
@@ -147,7 +147,7 @@ async def app_fixture(
     application = await model.deploy(
         charm, resources=resources, application_name=app_name, config=app_config, series="focal"
     )
-    await model.wait_for_idle()
+    await model.wait_for_idle(raise_on_error=False)
```
jdkandersson, line 150: Curios what the reason for this is please? If an error occurs can the tests keep going?

src/charm.py
```diff
@@ -387,23 +387,23 @@ def _set_up_discourse(self, event: HookEvent) -> None:
...
             self.model.unit.status = MaintenanceStatus("Compiling assets")
-            precompile_process: ExecProcess = container.exec(
-                [f"{DISCOURSE_PATH}/bin/bundle", "exec", "rake", "assets:precompile"],
-                environment=env_settings,
-                working_dir=DISCOURSE_PATH,
-                user="discourse",
-            )
-            precompile_process.wait_output()
+            try:
+                precompile_process: ExecProcess = container.exec(
```
jdkandersson, line 394: Perhaps for a different PR, I remember someone mentioned that precompile isn't strictly necessary, if the assets aren't precompiled that will happen at runtime. Something to think about is to do this after going into active so that the startup time is sped up

rockcraft.yaml
```diff
@@ -0,0 +1,203 @@
...
+      ".*": srv/discourse/app/
+  get-solved-plugin-source:
+    plugin: dump
+    source: https://github.com/discourse/discourse-solved.git
+    source-depth: 1
+    source-commit: d6c8089ca38611b09a8edb29d64f359bcef11f11
+    after:
+      - get-discourse
+    organize:
+      "*": srv/discourse/app/plugins/discourse-solved/
```
jdkandersson, line 76: Perhaps for when the rockfile is more stable, it would be good to include some documentation with reasoning for the different parts here, e.g., why installing from git etc

rockcraft.yaml
```diff
@@ -0,0 +1,203 @@
...
+    override-build: |
+      curl -fsSL https://deb.nodesource.com/setup_18.x | bash -
+      apt-get install -y nodejs
+      node --version
+      npm --version
+
+      npm install -g terser uglify-js yarn
+      yarn --cwd ${CRAFT_STAGE}/srv/discourse/app install --production --frozen-lockfile
+      yarn --cwd ${CRAFT_STAGE}/srv/discourse/app cache clean
+  get-gems:
```
jdkandersson, line 150: Is the `get-*` recommended naming? I think these are parts, i.e., components so perhaps just `gems` or `discourse` etc is enough?

nrobinaubertin: I like the idea that the name of the parts (with the `after` property) could give the whole story of the build.
For this rockfile it would be:
```
- install-node
- install-runtime-packages
- get-discourse (after install-node)
- get-solved-plugin-source (after get-discourse)
- get-markdown-note-plugin-source (after get-discourse)
- get-mermaid-plugin-source (after get-discourse)
- get-saml-plugin-source (after get-discourse)
- get-patches
- apply-patches (after get-patches and get-discourse)
- install-gems (after get-discourse, get-saml-plugin-source and apply-patches)
- get-scripts
```
I feel like it reads itself pretty well. Also parts are more "actions" than "things"

tests/integration/conftest.py
nrobinaubertin: I'll remove this later but as a note: it can happen that some charms reach an error state and recover soon after. This happens for the redis-k8s charm for example. So this could lead to us retrying CIs without needing it.

src/charm.py
nrobinaubertin: Do you have a ref for this so that I can create the corresponding issue ?

tests/integration/conftest.py
jdkandersson: Perhaps worth giving that feedback to the owner of the redis charm? It doesn't sound like the charm should enter error state

src/charm.py
jdkandersson: No reference, perhaps this would be a spike to investigate whether it is possible to skip the precompile action and the tests still pass? And see what the impact is to the user experience

rockcraft.yaml
jdkandersson: The snapcraft documentation includes a definition of a part which states that it is a component: https://snapcraft.io/docs/glossary

merkata: This is a good discussion on naming, I like it. The lines here do tell a story and I like it. Looking at the glossary I interpret it that a `part` is more of a component combined with instructions on how to fetch and build it and they are reusable, so I'm leaning more towards "actions" as well. If we treat them as components and name them as such, perhaps the parts above need to be combined to reflect the glossary and the idea that fetch and build is included in a component, so:

```text
- get-solved-plugin-source
- get-markdown-note-plugin-source
- get-mermaid-plugin-source
- ...
```
will result in something like
```text
- discourse-plugins
```
with the implication that they will be fetched and built in that part...
Just a thought on my end.

tests/integration/conftest.py
nrobinaubertin: I propose to handle that with https://warthogs.atlassian.net/browse/ISD-1001.

src/charm.py
nrobinaubertin: I created a spike for this: https://warthogs.atlassian.net/browse/ISD-1001

src/charm.py
```diff
@@ -330,7 +329,7 @@ def _create_discourse_environment_settings(self) -> typing.Dict[str, str]:
             # I need to take the required envVars for the application to work properly
             "CONTAINER_APP_NAME": "discourse",
             "CONTAINER_APP_ROOT": "/srv/discourse",
-            "CONTAINER_APP_USERNAME": "discourse",
+            "CONTAINER_APP_USERNAME": "root",
```
merkata, line 332: Is this due to rocks defaulting to root? There is the possibility to add a non-root user now as [annonunced](https://discourse.ubuntu.com/t/unifying-user-identity-across-snaps-and-rocks/36469/2) and described [here](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/reference/rockcraft.yaml/#run-user). It would be “_daemon_” (with UID/GID 584792), so might take some adjustments, though better than root in the long run. This can be done in a different PR, we just need to keep in the backlog.

merkata (dismissed):

src/charm.py
weiiwang01: Second this, others look good to me, thanks!

weiiwang01 (dismissed):

rockcraft.yaml
nrobinaubertin: I added some comments.  
As for why installing from git: because the dockerfile used to do it that way (I didn't want to change things if not needed) and I'm not even sure there's another way to get discourse ?

rockcraft.yaml
jdkandersson: any thoughts @gregory-schiano? Based on our interactions with the starcraft team and the docs it sounds like the intent for parts is not really actions (though they define actions to make a part available). So a part should be called `discourse` if it installs discourse and `dependencies` if it install dependencies etc

.github/workflows/integration_test_with_secrets.yaml
```diff
@@ -4,40 +4,15 @@ on:
...
-          docker push localhost:32000/discourse:test
-
-      - name: Run integration tests
-        run: |
-          tox -e integration -- \
-          -m requires_secrets \
-          --saml-email ${{ secrets.TEST_SAML_EMAIL }} \
-          --saml-password ${{ secrets.TEST_SAML_PASSWORD }} \
-          --discourse-image=localhost:32000/discourse:test
+  integration-tests-with-secrets:
```
arturo-seijas, line 7: We can probably remove this wf and use the "integration test" one without the `-m` argument, right?

nrobinaubertin: Sadly we can't. There are two integ tests conflicting with each other right now. I need to make sure which one are and fix it but this will be for another PR. I'll create a ticket for that.

weiiwang01 (approved):

arturo-seijas (approved):
LGTM

PR #76, opened 2023-07-05
Pin ops to version 2.3.0

arturo-seijas:
Pin ops to version 2.3.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-ops
+ops==2.3.0
 ops-lib-pgsql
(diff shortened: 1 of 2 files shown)


yanksyoon (approved):
Out of curiosity, what is the reason for the pin? Is there a bug with later versions?

arturo-seijas:
> Out of curiosity, what is the reason for the pin? Is there a bug with later versions?

There's a breaking change. Will update the dependency later

amandahla (approved):
LGTM

PR #79, opened 2023-07-11
Use the new database interface for the postgresql relation

nrobinaubertin:
I created a database.py module alongside the charm.  
It uses the new [postgresql_client](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/postgresql_client/v0/README.md) interface.


src/charm.py
```diff
@@ -359,14 +372,25 @@ def _are_db_relations_ready(self) -> bool:
...
-        if not self._stored.db_name:  # type: ignore
+        db_rel = self._database.get_relation_data()
+        if db_rel is None:
             self.model.unit.status = WaitingStatus("Waiting for database relation")
             return False
+        if None in (
+            db_rel["POSTGRES_USER"],
+            db_rel["POSTGRES_PASSWORD"],
+            db_rel["POSTGRES_DB"],
+            db_rel["POSTGRES_HOST"],
```
mthaddon, line 383: `POSTGRES_HOST` is never going to be `None`, btw, since you default to `":".split(":")[0]`.

nrobinaubertin: Should I leave it like this, as defensive code (against eventual changes in src/database.py) ?

mthaddon: I'd prefer to see this check handled in `src/database.py` directly, but if we're leaving it here for now, I think we should update this to say if the other values are `None` or if this one is `''`.

metadata.yaml
```diff
@@ -43,9 +43,9 @@ requires:
   redis:
     interface: redis
     limit: 1
-  db:
-    interface: pgsql
-    limit: 1
+  database:
+      interface: postgresql_client
+      limit: 1
```
mthaddon, line 48: Any reason for the extra spaces before `interface` and `limit` here? Would be good to consistent with existing entries

nrobinaubertin: Fixed

src/charm.py
nrobinaubertin: I've moved most of the related logic to the database module (and improved the consistency of the get_relation_data() function).

src/charm.py
```diff
@@ -72,11 +74,20 @@ def __init__(self, *args):
         """Initialize defaults and event handlers."""
         super().__init__(*args)
 
+        self._database = DatabaseObserver(self)
+
+        self.framework.observe(
+            self._database.database.on.database_created, self._database_relation_changed
```
jdkandersson, line 80: Is this using the observer pattern for database? If so, what is the reason for putting the event handlers here? Also, the name of the event handling functions should match the name of the event

jdkandersson: Also applicable to the other event handlers being added

src/charm.py
```diff
@@ -245,22 +251,31 @@ def _get_s3_env(self) -> typing.Dict[str, typing.Any]:
...
 
         Returns:
-            Dictionary with all the environment settings.
+            Tuple with the hostname and port of the related redis
         """
-        # Get redis connection information from the relation.
-        redis_hostname = None
-        redis_port = 6379
         # This is the current recommended way of accessing the relation data.
         for redis_unit in self._stored.redis_relation:  # type: ignore
```
jdkandersson, line 261: Hmm this doesn't look great, I guess we should moved away from using `_stored`

src/charm.py
```diff
@@ -464,7 +481,7 @@ def _config_changed(self, event: HookEvent) -> None:
             self._config_force_https()
             self.model.unit.status = ActiveStatus()
 
-    def _reload_configuration(self) -> None:
+    def _reload_configuration(self, _=None) -> None:
```
jdkandersson, line 484: I'm guessing this is because this is now used as an event handler, as noted above each event should have its own dedicated event handling function

src/database.py
```diff
@@ -0,0 +1,78 @@
...
+import typing
+
+from charms.data_platform_libs.v0.data_interfaces import DatabaseRequires
+from ops.charm import CharmBase
+from ops.framework import Object
+
+DATABASE_NAME = "discourse"
+
+
+class DatabaseObserver(Object):
```
jdkandersson, line 15: Perhaps work with @gregory-schiano to clarify how to use observers? Or perhaps with @amandahla who has recently implemented it on Synapse: https://github.com/canonical/synapse-operator/blob/main/src/database_observer.py

src/charm.py
amandahla, line 80: Could you rename the methods according to the ISD014, please?

"In the __init__ of a CharmBase class, self.framework.observe is used to register functions that handle charm events. There should be a 1-1 mapping between events and observers, where the observer is named _on_<event-name> (for example, _on_config_changed)."

src/charm.py
```diff
@@ -72,11 +74,20 @@ def __init__(self, *args):
...
 
+        self._database = DatabaseObserver(self)
+
+        self.framework.observe(
+            self._database.database.on.database_created, self._database_relation_changed
+        )
+        self.framework.observe(
+            self._database.database.on.endpoints_changed, self._database_relation_changed
+        )
+        self.framework.observe(
```
amandahla, line 85: I believe that the database events are here for the same reason that I put them there on the first version of Synapse PR to add database integration, right? :)
Do you think is possible to use the same solution by passing pebble/container as a parameter?

src/charm.py
```diff
@@ -245,22 +251,31 @@ def _get_s3_env(self) -> typing.Dict[str, typing.Any]:
...
-        redis_hostname = None
-        redis_port = 6379
         # This is the current recommended way of accessing the relation data.
         for redis_unit in self._stored.redis_relation:  # type: ignore
-            redis_hostname = self._stored.redis_relation[redis_unit].get("hostname")  # type: ignore
-            redis_port = self._stored.redis_relation[redis_unit].get("port")  # type: ignore
+            # mypy fails to see that this is indexable
+            redis_unit_data = self._stored.redis_relation[redis_unit]  # type: ignore
+            redis_hostname = redis_unit_data.get("hostname", "")  # type: ignore
+            redis_port = redis_unit_data.get("port", 6379)  # type: ignore
```
amandahla, line 265: Should 6379 be a constant?

src/charm.py
```diff
@@ -245,22 +251,31 @@ def _get_s3_env(self) -> typing.Dict[str, typing.Any]:
...
             )
+        return (redis_hostname, redis_port)
+
+    def _create_discourse_environment_settings(self) -> typing.Dict[str, typing.Any]:
+        """Create a layer config based on our current configuration.
+
+        Returns:
+            Dictionary with all the environment settings.
+        """
+        database_relation_data = self._database.get_relation_data()
```
amandahla, line 277: Is guaranteed that for every _create_discourse_environment_settings call the get_relation_data is not None?

src/charm.py
nrobinaubertin: This is not using the observer pattern. I propose to revisit this at a later date because using the observer pattern here introduces a fair amount of code and the primary goal was to [deploy discourse in production ASAP](https://warthogs.atlassian.net/browse/ISD-481).  
I've already created [a story for that](https://warthogs.atlassian.net/browse/ISD-934) and it will be my priority after the deployments.

src/charm.py
nrobinaubertin: Moved to a constant

src/charm.py
nrobinaubertin: Renamed

src/charm.py
nrobinaubertin: I propose to do this sort of changes as part of [ISD-934](https://warthogs.atlassian.net/browse/ISD-934). See https://github.com/canonical/discourse-k8s-operator/pull/79#discussion_r1274056536.

src/database.py
nrobinaubertin: Answered with https://github.com/canonical/discourse-k8s-operator/pull/79#discussion_r1274056536

src/charm.py
nrobinaubertin: [We can't](https://github.com/canonical/redis-k8s-operator/issues/74) for now.

src/charm.py
nrobinaubertin: Removed with the event handler renaming

src/database.py
jdkandersson: It is a bit confusing to call this an observer if we don't intend to actually implement the pattern?

nrobinaubertin: Renamed to "handler"

src/charm.py
nrobinaubertin: When I wrote this initially, I thought that it was covered by the numerous calls to `_are_db_relations_ready()`. That function makes sure using the `_is_relation_ready()` from the database module that the relation is OK.  

But it's actually not the case.  
Here is the calling tree of `_create_discourse_environment_settings()`:
```
_create_discourse_environment_settings
    _config_force_https
        _config_changed [OK]
            __init__
    _create_layer_config
        _reload_configuration [*]
            _config_changed [OK]
                __init__
            _on_database_relation_broken [KO]
                __init__
            _on_database_endpoints_changed [OK]
                __init__
            _on_database_created [OK]
                __init__
            _on_redis_relation_changed [OK]
                __init__
    _on_add_admin_user_action [KO]
        __init__
    _on_anonymize_user_action [KO]
        __init__
    _set_up_discourse [OK]
        __init__
```

When there's `[OK]` it means that there's a `_are_db_relations_ready()` check. Otherwise it's `[KO]`.  
The actions would be easy to fix.  
What comes after `_reload_configuration()` is subject to `_is_setup_completed()`. The only issues is with `_on_database_relation_broken()`.  
When we brake the relation, we want to update the config even if `get_relation_data()` returns `None`.  

So I decided that the correct thing to do is to always return a well formed `dict[str, str]` with `get_relation_data()` and to change `is_relation_ready()` accordingly. The idea is that we should be able to change the config/layer's discourse at *any* time. The charm has to adapt to these changes.

requirements.txt
```diff
@@ -1,2 +1 @@
-ops
```
yanksyoon, line 1: May I know why the ops lib is being removed?

src/charm.py
```diff
@@ -245,22 +282,31 @@ def _get_s3_env(self) -> typing.Dict[str, typing.Any]:
 
         return s3_env
 
-    def _create_discourse_environment_settings(self) -> typing.Dict[str, typing.Any]:
-        """Create a layer config based on our current configuration.
+    def _get_redis_relation_data(self) -> typing.Tuple[typing.Any, typing.Any]:
```
yanksyoon, line 285: Nitpick: May I suggest a stricter return type hints here if possible?

src/charm.py
amandahla: For this PR is a nice solution.
I think this might be more clear/revised when working with the Observer pattern so you can even create an exception for this case.

amandahla (dismissed):
LGTM

requirements.txt
arturo-seijas, line 1: No ops?

src/charm.py
```diff
@@ -245,22 +282,31 @@ def _get_s3_env(self) -> typing.Dict[str, typing.Any]:
...
+            # mypy fails to see that this is indexable
+            redis_unit_data = self._stored.redis_relation[redis_unit]  # type: ignore
+            redis_hostname = redis_unit_data.get("hostname", "")  # type: ignore
+            redis_port = redis_unit_data.get("port", DEFAULT_REDIS_PORT)  # type: ignore
             logger.debug(
                 "Got redis connection details from relation of %s:%s", redis_hostname, redis_port
             )
+        return (redis_hostname, redis_port)
+
+    def _create_discourse_environment_settings(self) -> typing.Dict[str, typing.Any]:
```
yanksyoon, line 302: Nitpick:
I'm wondering if the `database_relation_data` and `redis_relation_data` would be better instantiated on the handler and be passed to environment settings as variables. I think it could help understand the dataflow better.
I understand that this could be a stylistic preference though.

src/database.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        super().__init__(charm, "database-observer")
+        self._charm = charm
+        self.database = DatabaseRequires(
+            self._charm,
+            relation_name=self._RELATION_NAME,
+            database_name=DATABASE_NAME,
+        )
+
+    def get_relation_data(self) -> typing.Dict[str, str]:
```
yanksyoon, line 34: Nitpick:
I'm wondering if we introduce a typed dataclass it would be easier to digest the code :D, I usually code with `tab` keys on my IDE 😂

tests/integration/conftest.py
```diff
@@ -160,13 +161,23 @@ async def app_fixture(
             config=app_config,
             series="focal",
         )
+
+    await model.wait_for_idle()
```
arturo-seijas, line 165: Perhaps specifying the status to wait for can be useful to debug when this fails

tests/integration/conftest.py
```diff
@@ -160,13 +161,23 @@ async def app_fixture(
...
+
+    # configure postgres
+    await postgres_app.set_config(
+        {
+            "plugin_hstore_enable": "true",
+            "plugin_pg_trgm_enable": "true",
+        }
+    )
     await model.wait_for_idle()
```
arturo-seijas, line 174: Same here

tests/integration/conftest.py
```diff
@@ -160,13 +161,23 @@ async def app_fixture(
...
 
     # Add required relations
     unit = model.applications[app_name].units[0]
     assert unit.workload_status == WaitingStatus.name  # type: ignore
     await asyncio.gather(
-        model.add_relation(app_name, "postgresql-k8s:db-admin"),
+        model.add_relation(app_name, "postgresql-k8s:database"),
```
arturo-seijas, line 180: I'd suggest you switch to the latest/edge redis, as the revision in stable has some stability issues that might cause flakiness

tests/unit/test_charm.py
```diff
@@ -569,15 +562,24 @@ def test_install_when_not_leader(self):
...
         "Add postgresql relation and relation data to the charm."
-        self.harness.charm._stored.db_name = "discourse-k8s"
-        self.harness.charm._stored.db_user = "someuser"
-        self.harness.charm._stored.db_password = "somepasswd"  # nosec
-        self.harness.charm._stored.db_host = "dbhost"
+
+        relation_data = {
+            "database": DATABASE_NAME,
+            "endpoints": "dbhost:5432,dbhost-2:5432",
+            "password": "somepasswd",  # nosec
```
yanksyoon, line 569: Being a nitpicker here, may I suggest secrets.token_hex(16)?

requirements.txt
nrobinaubertin: It comes included with charmcraft now (and charmcraft complains if we have it in the requirements.txt)

yanksyoon (commented):
Looks great! Minor nitpicks as comments, would it be correct to assume that the stylistic issues are pretty hard to overcome at this stage of the charm?

requirements.txt
yanksyoon: Just saw the comment above, resolving! TIIL!

arturo-seijas: Nice!

nrobinaubertin: It seems to come included with charmcraft now (and charmcraft complains if we have it in the requirements.txt)

src/database.py
nrobinaubertin: Text editor here, every character is typed !

More seriously, `get_relation_data` is consumed only twice, is it worth it to introduce a dataclass ?

src/charm.py
nrobinaubertin: I want to leave what is related to redis in the charm (for the scope of this PR at least), since the database module will be about postgres only.

It's `get_relation_data()` that creates `database_relation_data`, so it's already in the database module. I'm not sure to follow what you mean

src/charm.py
nrobinaubertin: Fixed

tests/unit/test_charm.py
nrobinaubertin: Fixed

src/database.py
```diff
@@ -0,0 +1,86 @@
...
+from ops.charm import CharmBase
+from ops.framework import Object
+
+DATABASE_NAME = "discourse"
+
+
+class DatabaseHandler(Object):
+    """The Database relation observer."""
+
+    _RELATION_NAME = "database"
```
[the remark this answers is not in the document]
nrobinaubertin: I renamed it and moved it to the charm module

tests/integration/conftest.py
nrobinaubertin: status added

tests/integration/conftest.py
nrobinaubertin: status added

tests/integration/conftest.py
nrobinaubertin: Switched to latest/edge

src/charm.py
yanksyoon: I was thinking instead of 
```
def _create_discourse_environment_settings(self)
    database_relation_data = self._database.get_relation_data()
    redis_relation_data = self._get_redis_relation_data()
```,
it would be better to do 
```
def _create_discourse_environment_settings(self, db_relation_data, redis_relation_data)
    ...
```
so that the dataflow is singular rather than being referred / called in different places within the execution flow. However, it's a minor stylistic difference. Would this explanation be a bit clearer?
Btw feel free to resolve this, I don't think it's important change within this scope! :D

src/database.py
yanksyoon: I see! Probably not :D

src/charm.py
nrobinaubertin: I see.  
But here, it's the only place in the whole charm where I refer to get_relation_data(), so it's already singular (and it would not be with your proposal, I would have to add it at 5 different places).

So I'm going to leave it as is

mthaddon (approved):
LGTM - we'll want to update the docs to direct people to use this new relation and charm, but that can happen as a later step.

amandahla (approved):
LGTM

PR #83, opened 2023-08-09
Fix _are_db_relations_ready() and _get_redis_relation_data()

nrobinaubertin:
- `_get_redis_relation_data()` should *always* return a Tuple[str, int]
- `_are_db_relations_ready()` should return `False` on malformed redis relation data

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -66,7 +66,6 @@ SCRIPT_PATH = "/srv/scripts"
 SERVICE_NAME = "discourse"
 SERVICE_PORT = 3000
 SETUP_COMPLETED_FLAG_FILE = "/run/discourse-k8s-operator/setup_completed"
-DEFAULT_REDIS_PORT = 6379
 DATABASE_RELATION_NAME = "database"
 
 
@@ -293,8 +292,15 @@ class DiscourseCharm(CharmBase):
         for redis_unit in self._stored.redis_relation:  # type: ignore
             # mypy fails to see that this is indexable
             redis_unit_data = self._stored.redis_relation[redis_unit]  # type: ignore
-            redis_hostname = redis_unit_data.get("hostname", "")  # type: ignore
-            redis_port = redis_unit_data.get("port", DEFAULT_REDIS_PORT)  # type: ignore
+            try:
+                redis_hostname = str(redis_unit_data.get("hostname"))  # type: ignore
+            # I need to catch all exceptions that str() can throw
+            except Exception:  # pylint: disable=broad-exception-caught
+                redis_hostname = ""
+            try:
+                redis_port = int(redis_unit_data.get("port"))  # type: ignore
+            except ValueError:
+                redis_port = 0
             logger.debug(
                 "Got redis connection details from relation of %s:%s", redis_hostname, redis_port
             )
@@ -416,7 +422,10 @@ class DiscourseCharm(CharmBase):
         if not self._stored.redis_relation:  # type: ignore
             self.model.unit.status = WaitingStatus("Waiting for redis relation")
             return False
-        if self._get_redis_relation_data()[0] == "":
+        if (
+            self._get_redis_relation_data()[0] in ("", "None")
+            or self._get_redis_relation_data()[1] == 0
+        ):
             self.model.unit.status = WaitingStatus("Waiting for redis relation to initialize")
             return False
         return True
--- a/src/charm.py
+++ b/src/charm.py
@@ -415,10 +415,7 @@ class DiscourseCharm(CharmBase):
         if not self._stored.redis_relation:  # type: ignore
             self.model.unit.status = WaitingStatus("Waiting for redis relation")
             return False
-        if (
-            self._get_redis_relation_data()[0] in ("", "None")
-            or self._get_redis_relation_data()[1] == 0
-        ):
+        if not all(self._get_redis_relation_data()):
             self.model.unit.status = WaitingStatus("Waiting for redis relation to initialize")
             return False
         return True
--- a/src/charm.py
+++ b/src/charm.py
@@ -292,16 +292,8 @@ class DiscourseCharm(CharmBase):
         for redis_unit in self._stored.redis_relation:  # type: ignore
             # mypy fails to see that this is indexable
             redis_unit_data = self._stored.redis_relation[redis_unit]  # type: ignore
-            try:
-                redis_hostname = str(redis_unit_data.get("hostname"))  # type: ignore
-            # I need to catch all exceptions that str() can throw
-            except Exception:  # pylint: disable=broad-exception-caught
-                redis_hostname = ""
-            try:
-                redis_port = int(redis_unit_data.get("port"))  # type: ignore
-            # I need to catch all exceptions that int() can throw
-            except Exception:  # pylint: disable=broad-exception-caught
-                redis_port = 0
+            redis_hostname = str(redis_unit_data.get("hostname", "") or "")
+            redis_port = int(redis_unit_data.get("port", 0) or 0)
             logger.debug(
                 "Got redis connection details from relation of %s:%s", redis_hostname, redis_port
             )
--- a/src/charm.py
+++ b/src/charm.py
@@ -292,8 +292,8 @@ class DiscourseCharm(CharmBase):
         for redis_unit in self._stored.redis_relation:  # type: ignore
             # mypy fails to see that this is indexable
             redis_unit_data = self._stored.redis_relation[redis_unit]  # type: ignore
-            redis_hostname = str(redis_unit_data.get("hostname", "") or "")
-            redis_port = int(redis_unit_data.get("port", 0) or 0)
+            redis_hostname = str(redis_unit_data.get("hostname"))
+            redis_port = int(redis_unit_data.get("port"))
             logger.debug(
                 "Got redis connection details from relation of %s:%s", redis_hostname, redis_port
             )
@@ -306,6 +306,9 @@ class DiscourseCharm(CharmBase):
             Dictionary with all the environment settings.
         """
         database_relation_data = self._database.get_relation_data()
+
+        # The following could fail if the data is malformed.
+        # We/don't catch it because we don't want to silently fail in those cases
         redis_relation_data = self._get_redis_relation_data()
 
         pod_config = {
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -293,8 +292,15 @@ def _get_redis_relation_data(self) -> typing.Tuple[str, int]:
         for redis_unit in self._stored.redis_relation:  # type: ignore
             # mypy fails to see that this is indexable
             redis_unit_data = self._stored.redis_relation[redis_unit]  # type: ignore
-            redis_hostname = redis_unit_data.get("hostname", "")  # type: ignore
-            redis_port = redis_unit_data.get("port", DEFAULT_REDIS_PORT)  # type: ignore
+            try:
+                redis_hostname = str(redis_unit_data.get("hostname"))  # type: ignore
```
mthaddon, line 296: How about just `redis_hostname = str(redis_unit_data.get("hostname") or "")`? This would ensure it's either a valid value or "" I think?

src/charm.py
```diff
@@ -293,8 +292,15 @@ def _get_redis_relation_data(self) -> typing.Tuple[str, int]:
...
+            # I need to catch all exceptions that str() can throw
+            except Exception:  # pylint: disable=broad-exception-caught
+                redis_hostname = ""
```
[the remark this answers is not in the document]
nrobinaubertin: I decided that `_get_relation_data()` should only make sure that the output is Tuple[str, int].  
I handle malformed relations in `_are_db_relations_ready()`

src/charm.py
gregory-schiano: I don't think catching `Exception` is a great solution.
What do you want to protect against ?
I'm wondering if an earlier pydantic validation of the data isn't a better approach

nrobinaubertin: The redis lib is happy to put whatever in the stored state: https://github.com/canonical/redis-k8s-operator/blob/main/lib/charms/redis_k8s/v0/redis.py#L79

Do we know the interface for the databag elements ?

weiiwang01: Juju relation data values are always strings. I think it's okay to just use `redis_unit_data.get("hostname", "")`.

src/charm.py
```diff
@@ -293,8 +292,16 @@ def _get_redis_relation_data(self) -> typing.Tuple[str, int]:
...
+            try:
+                redis_port = int(redis_unit_data.get("port"))  # type: ignore
+            # I need to catch all exceptions that int() can throw
+            except Exception:  # pylint: disable=broad-exception-caught
```
weiiwang01, line 303: I think `int` will only throw `TypeError` or `ValueError` since the relation data are always string typed, so the value of `redis_unit_data.get("port")` will either be a string, or `None`.

weiiwang01: Actually I have a new idea, we can
```python
redis_port_str = redis_unit_data.get("port", "")
redis_port = int(redis_port_str) if redis_port_str.isnumeric() else 0
```

src/charm.py
weiiwang01: Yes, you are right, sorry I have misread that, I thought it's reading from the relation.

src/charm.py
weiiwang01: Sorry I was wrong about this, we also need to consider the situation of `redis_unit_data.get("port", "")` being `None`.

src/charm.py
```diff
@@ -416,7 +423,10 @@ def _are_db_relations_ready(self) -> bool:
         if not self._stored.redis_relation:  # type: ignore
             self.model.unit.status = WaitingStatus("Waiting for redis relation")
             return False
-        if self._get_redis_relation_data()[0] == "":
+        if (
+            self._get_redis_relation_data()[0] in ("", "None")
+            or self._get_redis_relation_data()[1] == 0
```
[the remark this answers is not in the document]
nrobinaubertin: Changed

src/charm.py
nrobinaubertin: Outdated

weiiwang01:
```python3
redis_port_data = redis_unit_data.get("port")
redis_port = int(redis_port_str) if redis_port_data and redis_port_data.isnumeric() else 0
```

weiiwang01 (dismissed):

src/charm.py
```diff
@@ -293,8 +292,8 @@ def _get_redis_relation_data(self) -> typing.Tuple[str, int]:
         for redis_unit in self._stored.redis_relation:  # type: ignore
             # mypy fails to see that this is indexable
             redis_unit_data = self._stored.redis_relation[redis_unit]  # type: ignore
-            redis_hostname = redis_unit_data.get("hostname", "")  # type: ignore
-            redis_port = redis_unit_data.get("port", DEFAULT_REDIS_PORT)  # type: ignore
+            redis_hostname = str(redis_unit_data.get("hostname", "") or "")
```
jdkandersson, line 295: Should this raise an error if the expected data is not present? Returning an empty string when it isn't could be misleading to the caller. Same for the port

tests/unit/test_charm.py
```diff
@@ -654,3 +654,51 @@ def test_postgres_relation_data(self):
...
+                {"hostname": "redis-port"},
+                False,
+            ),
+        ]
+
+        for relation_data, should_be_ready in test_cases:
+            with self.subTest(relation_data=relation_data, should_be_ready=should_be_ready):
+                self.start_harness(with_postgres=True, with_redis=False)
+                # get a relation ID for the test outside of __init__ (note pylint disable)
+                self.redis_relation_id = (  # pylint: disable=attribute-defined-outside-init
```
jdkandersson, line 699: What is the reason for needing to set this? Would a local variable work?

src/charm.py
mthaddon: This function is always returning a tuple of a string and an integer for hostname and port. Later on in the code we're validating if this sufficient to consider the relation "ready" as follows: `if not all(self._get_redis_relation_data()):`. Is there a particular reason you think it's better to validate the data here/raise an error if it's not complete?

tests/unit/test_charm.py
```diff
@@ -654,3 +654,51 @@ def test_postgres_relation_data(self):
...
+                False,
+            ),
+            (
+                {"port": 6379},
+                False,
+            ),
+            (
+                {"hostname": "redis-port"},
+                False,
+            ),
```
mthaddon, line 692: I think you're missing test cases where the port is 0, but you're doing `int(redis_unit_data.get("port", 0) or 0)` so it could very well be 0. I think it'll work fine, but would be good to include in the test cases.

src/charm.py
jdkandersson: yes because the data missing is essentially an error which can be caught with a `try ... except`. This way, it is explicit that something is wrong rather than implicit based on the value

tests/unit/test_charm.py
nrobinaubertin: Fixed

src/charm.py
nrobinaubertin: I changed the code to test if errors are raised outside of `get_redis_relation_data()` (following David's proposal).  
I decided to not do those checks in `_create_discourse_environment_settings()` because the data *should* be okay there (if it's not, then we end in error state but that is to be expected anyway).

tests/unit/test_charm.py
nrobinaubertin: I added a test case with port 0

src/charm.py
nrobinaubertin: I also added a custom exception for the cases when a ValueError|TypeError is raised after discussion with David.

src/charm.py
```diff
@@ -416,8 +428,14 @@ def _are_db_relations_ready(self) -> bool:
         if not self._stored.redis_relation:  # type: ignore
             self.model.unit.status = WaitingStatus("Waiting for redis relation")
             return False
-        if self._get_redis_relation_data()[0] == "":
-            self.model.unit.status = WaitingStatus("Waiting for redis relation to initialize")
+        try:
+            if (
+                self._get_redis_relation_data()[0] in ("", "None")
+                or self._get_redis_relation_data()[1] == 0
+            ):
```
mthaddon, line 435: I'm not sure you need this now do you? You just need to run `_get_redis_relation_data()` and if it doesn't raise an error you know the data is good?

nrobinaubertin: If the redis port is None, the str(None) will not raise an exception. Same for a 0 port. Do we want to support port 0 ?

src/charm.py
```diff
@@ -307,6 +316,9 @@ def _create_discourse_environment_settings(self) -> typing.Dict[str, str]:
             Dictionary with all the environment settings.
         """
         database_relation_data = self._database.get_relation_data()
+
+        # The following could fail if the data is malformed.
+        # We/don't catch it because we don't want to silently fail in those cases
```
jdkandersson, line 321: I guess in that case the charm goes into ErrorState

jdkandersson (dismissed):

src/charm.py
jdkandersson: To be clear, this was an observation trying to prompt you to think whether that is as intended 🙂

src/charm.py
mthaddon: Ok, this seems fine for now, and I think should later be refactored into a `redis`  module similar to the `database` module, but that can/should be handled separately (we should use a similar pattern for both).

src/charm.py
```diff
@@ -416,8 +428,14 @@ def _are_db_relations_ready(self) -> bool:
...
-            self.model.unit.status = WaitingStatus("Waiting for redis relation to initialize")
+        try:
+            if (
+                self._get_redis_relation_data()[0] in ("", "None")
+                or self._get_redis_relation_data()[1] == 0
+            ):
+                self.model.unit.status = WaitingStatus("Waiting for redis relation to initialize")
+                return False
+        except MissingRedisRelationDataError:
             return False
```
mthaddon, line 439: I think we'd want to set some kind of status on the unit here wouldn't we?

src/charm.py
nrobinaubertin: Yes, it's exactly what we would want in that situation.

src/charm.py
nrobinaubertin: Fixed

mthaddon (approved):
Looks good, thx

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #92, opened 2023-08-17
Force S3 uploads

nrobinaubertin:
This adds a workaround for assets being skipped during the s3:upload_assets process.  
It should be revisited later when upgrading to a new version of discourse.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -511,6 +511,10 @@ class DiscourseCharm(CharmBase):
                 else "",
             )
         env_settings = self._create_discourse_environment_settings()
+        if self.config.get("s3_enabled"):
+            # We force assets to be uploaded to S3
+            # This should be considered as a workaround and revisited later
+            env_settings.update({"FORCE_S3_UPLOADS": "true"})
         try:
             if self.model.unit.is_leader() and self._should_run_s3_migration(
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -511,6 +511,10 @@ def _config_changed(self, event: HookEvent) -> None:
                 else "",
             )
         env_settings = self._create_discourse_environment_settings()
+        if self.config.get("s3_enabled"):
+            # We force assets to be uploaded to S3
+            # This should be considered as a workaround and revisited later
+            env_settings.update({"FORCE_S3_UPLOADS": "true"})
```
mthaddon, line 517: I think this might be better in the `_get_s3_env` method?

nrobinaubertin: I wanted to force this only when uploading assets that way. To avoid further possible disruptions by this workaround

nrobinaubertin: But I'm okay to change its location if you think it's better there

nrobinaubertin: I went ahead and moved it to `_get_s3_env` as it is conceptually simpler

mthaddon (approved):

arturo-seijas (approved):
LGTM

PR #105, opened 2023-09-15
Use pytest for the unit tests

nrobinaubertin:
### Overview

Use pytest for the unit tests

### Checklist


tests/unit/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+
+import ops
+import pytest
+from ops.model import BlockedStatus, Container
+from ops.testing import Harness
+
+from tests.unit import helpers
+from tests.unit._patched_charm import DiscourseCharm, pgsql_patch
+
+BLOCKED_STATUS = BlockedStatus.name  # type: ignore
```
jdkandersson, line 18: @benhoyt this shouldn't be needed anymore, right? Or is that change not released yet?

tests/unit/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+
+DATABASE_NAME = "discourse"
+
+
+@pytest.fixture(name="start_harness", autouse=True)
+def start_harness_fixture():  # -> typing.Generator[Harness, None, None]:
```
jdkandersson, line 24: Should the type hint either be uncommented or removed?

tests/unit/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+    """Ops testing framework harness fixture."""
+
+    def start_harness(
```
jdkandersson, line 27: What is the reason for making this a fixture rather than a helper function please?

tests/unit/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+        doesn't reinitialize the charm as expected.
+        Ref: https://github.com/canonical/operator/issues/736
+
+        Args:
+            - with_postgres: should a postgres relation be added
+            - with_redis: should a redis relation be added
+        """
+        harness = Harness(DiscourseCharm)
+        harness.disable_hooks()
+        # Protected access is needed to instantiate harness
```
jdkandersson, line 46: This seems to warrant a more detailed comment, I would be surprised if it is intentional to use the private harness methods to instantiate the harness

tests/unit/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+            with helpers.patch_exec():
+                harness.container_pebble_ready("discourse")
+
+        return harness
+
+    return start_harness
+
+
+@contextlib.contextmanager
+def _patch_setup_completed():
```
jdkandersson, line 83: Don't we have a new way of doing this in tests with being able to interact with the file system of the container?

tests/unit/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+    # pylint: disable=protected-access
+    harness.charm._stored.redis_relation = {
+        redis_relation_id: {"hostname": "redis-host", "port": 1010}
+    }
+
+
+def _add_ingress_relation(harness):
+    "Add ingress relation and relation data to the charm."
+    nginx_route_relation_id = harness.add_relation("nginx-route", "ingress")
+    harness.add_relation_unit(nginx_route_relation_id, "ingress/0")
```
jdkandersson, line 134: Shouldn't these be fixtures?

tests/unit/helpers.py
```diff
@@ -0,0 +1,30 @@
...
+
+import contextlib
+import typing
+import unittest.mock
+
+import ops
+
+
+@contextlib.contextmanager
+def patch_exec(fail: bool = False) -> typing.Generator[unittest.mock.Mock, None, None]:
```
jdkandersson, line 14: Isn't there a new way of doing this now where we don't have to mock ourselves?

jdkandersson: Check with @weiiwang01 who helped develop it

tests/unit/test_charm.py
```diff
@@ -6,452 +6,335 @@
...
-        """
-        arrange: given a deployed discourse charm with all the required relations
-        act: when s3_enabled configuration is True and there's no s3_bucket
-        assert: it will get to blocked status waiting for the latter.
-        """
-        self.harness.update_config(
+@pytest.mark.parametrize(
+    "with_postgres, with_redis, with_ingress, status",
+    [
+        (False, False, False, WaitingStatus("Waiting for database relation")),
```
jdkandersson, line 28: Do you think it would be worth it to use `pytest.param` and use the `id` argument to describe what the case is?

tests/unit/test_charm.py
```diff
@@ -6,452 +6,335 @@
...
+    ],
+)
+def test_relations(start_harness, with_postgres, with_redis, with_ingress, status):
+    """
+    arrange: given a deployed discourse charm
+    act: when pebble ready event is triggered
+    assert: it will have the correct status depending on the relations
+    """
+    harness = start_harness(
+        with_postgres=with_postgres, with_redis=with_redis, with_ingress=with_ingress
```
jdkandersson, line 45: Do you think these should be fixtures rather than arguments to a function?

tests/unit/conftest.py
benhoyt: I'm not quite sure which change you're referring to. However, from a quick grep, it doesn't look like this value is actually used. It's defined separately in `tests/unit/test_charm.py` and used once there as follows:

```python
    assert harness.model.unit.status.name == BLOCKED_STATUS
```

However, I think the better and more Pythonic way to write that is to use `isinstance`:

```python
    assert isinstance(harness.model.unit.status, ops.BlockedStatus)
```

jdkandersson: Thanks Ben

Besides not being used, the reason this `type: ignore` used to be needed is because `BlockedStatus.name` and the `.name` attribute of other statuses used to be `None | str`, I believe it was `None` in the base class. I remember a PR that changed that

tests/unit/conftest.py
merkata: A link to an implementation would be helpful here.

jdkandersson: @weiiwang01 could you help please? I think this fixture uses both for the command execution and the file system: https://github.com/canonical/flask-k8s-operator/blob/609d60e723974f46269139dd2b1539e4b65fe592/tests/unit/conftest.py#L85

tests/unit/helpers.py
nrobinaubertin: I think that you're talking about [handle_exec](https://ops.readthedocs.io/en/latest/#ops.testing.Harness.handle_exec). But `handle_exec` gives us a way to intercept a call to do something else instead, not test that the call was made.  

I tried to use `handle_exec` with a handler that would raise an Exception, like that:
```
        def test_handler(args):
            raise Exception

        harness.handle_exec(
            "discourse", 
            [],
            handler=test_handler
        )
```

But the test didn't fail. So I'm not sure that I can use `handle_exec` here.

tests/unit/conftest.py
nrobinaubertin: Removed

tests/unit/conftest.py
nrobinaubertin: Changed to a helper function

tests/unit/conftest.py
nrobinaubertin: Removed

tests/unit/conftest.py
nrobinaubertin: They're meant to modify the object created by `start_harness` (which is now a helper function). Can they be fixtures with that in mind ?

tests/unit/test_charm.py
nrobinaubertin: Added

tests/unit/test_charm.py
nrobinaubertin: I think that this is the same conversation as [here](https://github.com/canonical/discourse-k8s-operator/pull/105#discussion_r1328251271)

tests/unit/conftest.py
nrobinaubertin: I rewrote this status check the way Ben proposed.

tests/unit/conftest.py
nrobinaubertin: I think that you're talking about the [get_filesystem_root](https://ops.readthedocs.io/en/latest/#ops.testing.Harness.get_filesystem_root) method.  
I tried to use it that way:
```
    def _set_setup_completed():
        root = harness.get_filesystem_root("discourse")
        (root / SETUP_COMPLETED_FLAG_FILE[1:]).parent.mkdir(parents=True, exist_ok=True)
        (root / SETUP_COMPLETED_FLAG_FILE[1:]).write_text("")
    harness.charm._set_setup_completed = _set_setup_completed
```

Before realizing that the tests didn't rely on it. So I removed that.

tests/unit/conftest.py
jdkandersson: Let's try this approach on here and see how we go. It is a little different to many of our charms that use fixtures to create a range of artefacts required for the test and then pick and choose which artefacts a test needs to be able to function

jdkandersson (dismissed):

tests/unit/helpers.py
jdkandersson: Perhaps discuss with @weiiwang01 to see if he has an idea of how to do this?

weiiwang01: Yes, we have talked, the reason might be that the current mocks intercepted the `ops.testing` calls.

tests/unit/helpers.py
```diff
@@ -0,0 +1,152 @@
...
+        exec_process_mock.wait_output = unittest.mock.MagicMock(return_value=("", ""))
+    else:
+        exec_process_mock.wait_output = unittest.mock.Mock()
+        exec_process_mock.wait_output.side_effect = ops.pebble.ExecError([], 1, "", "")
+    exec_function_mock = unittest.mock.MagicMock(return_value=exec_process_mock)
+    with unittest.mock.patch.multiple(ops.model.Container, exec=exec_function_mock):
+        yield exec_function_mock
+
+
+def start_harness(
```
amandahla, line 39: What an interesting approach! :clap: 

I tried to use fixtures to solve this problem of setting a harness with specific conditions for multiple tests (see discussion [here](https://github.com/canonical/synapse-operator/pull/19#discussion_r1302486670) in Synapse) but this, at least for me, it's clearer since you are using named parameters.

tests/unit/test_charm.py
amandahla: Just a reminder: Fixtures usage was considered hard to follow here -> https://github.com/canonical/synapse-operator/pull/19#discussion_r1302486670

amandahla (dismissed):
LGTM

tests/unit/helpers.py
nrobinaubertin: After discussing with Weii, we saw that I forgot to remove the patch to `Container.exec` during my test of `handle_exec`. This caused it to catch the exec call instead of `handle_exec`.  
I rewrote the tests using `handle_exec`.

jdkandersson (dismissed):

tests/unit/helpers.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # We catch all exec calls to the container by default
+    harness.handle_exec("discourse", [], result=0)
+
+    if with_postgres:
+        _add_postgres_relation(harness)
+
+    if with_redis:
+        _add_redis_relation(harness)
+        with patch_exec():
```
weiiwang01, line 71: I am a little curious why the `patch_exec` is still need here since the `harness.handle_exec` has called to handle everything above.

amandahla (dismissed):
LGTM

tests/unit/helpers.py
nrobinaubertin: Forgot to remove that. Fixed

jdkandersson (approved):

weiiwang01 (approved):

nrobinaubertin:
This PR is currently blocked by: https://github.com/canonical/discourse-k8s-operator/pull/106

PR #108, opened 2023-09-25
Force charm service to stop on db relation broken

nrobinaubertin:
### Overview

The charm stayed incorrectly in an active state when the db relation was broken.  
This PR makes sure that the service stops when this is the case.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -123,7 +123,7 @@ class DiscourseCharm(CharmBase):
         Args:
             event: Event triggering the database created handler.
         """
-        if self._are_db_relations_ready():
+        if self._are_relations_ready():
             self._reload_configuration()
 
     def _on_database_endpoints_changed(self, _: DatabaseEndpointsChangedEvent) -> None:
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -628,6 +632,17 @@ def _on_anonymize_user_action(self, event: ActionEvent) -> None:
                     f"Failed to anonymize user with username {username}:{ex.stdout}"  # type: ignore
                 )
 
+    def _stop_service(self):
+        """Stop discourse, this operation is idempotent."""
+        logger.info("Stopping discourse")
```
Thanhphan1147, line 637: Not a very big deal but I want to know if this logging line is intentional? Maybe we can add a bit more context to it?

Thanhphan1147 (approved):
I looked and it looks good to me!

jdkandersson (approved):

src/charm.py
nrobinaubertin: I found it interesting to put in the log that the service was stopped. As for the context, I'm not sure what could be added. This function is called just after a change of the charm's status that says why the charm is stopped

Thanhphan1147: Ah okay, thanks for clearing that up. It's ok for me then!

PR #111, opened 2023-09-27
Add _start_service() method and rework discourse setup

nrobinaubertin:
### Overview

This is a follow-up to this PR: https://github.com/canonical/discourse-k8s-operator/pull/108  
This fixes multiple things:
- Make sure that the service is restarted when the necessary relations are present (redis and postgres)
- Apply the db migrations when postgres gets related

### Checklist

- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)

#### Explanations for the checklist
- ISD054 will be addressed in the next cycle
- src-docs will be added in another PR

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -123,6 +123,8 @@ class DiscourseCharm(CharmBase):
         Args:
             event: Event triggering the database created handler.
         """
+        if self.unit.is_leader():
+            self._execute_migrations()
         if self._are_relations_ready():
             self._reload_configuration()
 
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -483,6 +500,43 @@ def _set_up_discourse(self, event: HookEvent) -> None:
...
+        """
+        container = self.unit.get_container(SERVICE_NAME)
+        if not self._are_relations_ready() or not container.can_connect():
+            event.defer()
+            return
+        try:
+            self._execute_migrations()
+            self._compile_assets()
+            self._set_setup_completed()
+            self._set_workload_version()
```
jdkandersson, line 539: The PR seems to be marked as compliant with ISD054 - this section doesn't look like it is since this seems to be business logic elated to dealing with discourse which generally should be in a separate discourse management module. I'm guessing in this case the reason for not refactoring is because either we need this for the migration or it is out of scope for the PR. Thats ok - in that case it would be good for the description to reflect this and include a reason for not being compliant please

jdkandersson (approved):

src/charm.py
nrobinaubertin: Yes, "managing charm complexity" is a goal for the next cycle. I agree that this already tends towards it. Let's say that I'm preparing for that at the same time.

jdkandersson: Great that we're working towards that and thanks for updating the description!

arturo-seijas (approved):
LGTM

PR #128, opened 2023-10-20
Add some pebble checks

nrobinaubertin:
### Overview

Add some pebble checks for the discourse service.

### Checklist


arturo-seijas (dismissed):
LGTM

amandahla (dismissed):
LGTM

src/charm.py
```diff
@@ -389,9 +389,22 @@ def _create_layer_config(self) -> ops.pebble.LayerDict:
...
+                "discourse-ready": {
                     "override": "replace",
-                    "http": {"url": f"http://localhost:{SERVICE_PORT}"},
+                    "level": "ready",
+                    "http": {"url": f"http://localhost:{SERVICE_PORT}/srv/status"},
+                },
+                "discourse-alive": {
+                    "override": "replace",
+                    "level": "alive",
+                    "tcp": {"port": SERVICE_PORT, "host": "localhost"},
```
mthaddon, line 400: I think we need to be very careful about adding any checks that are "alive" as I believe that means if they fail the container will be restarted. Are we sure that's what we want here? Will this be responding correctly during asset compilation or s3 uploads, for instance, or will this mean the container gets restarted mid-way through?

nrobinaubertin: From what I understand, this check will only have an effect after the 'ready' checks pass. That means that the 'ready' http check will be valid when this check kicks in (which already mean that the unicorn process has started).  
Failing this check doesn't mean that the container will be restarted from what I understand, but only the service ?

nrobinaubertin: After testing locally, I can reproduce the error of the CI, only when the alive check is there. So the issue is definitely coming from there.

nrobinaubertin: I removed the check

amandahla (approved):

gtrkiller (approved):
LGTM

PR #125, opened 2023-10-18
Pin plugins to specific commits

nrobinaubertin:
### Overview
- avoid surprise updates from transient dependencies or plugins
- add the data-explorer plugin
- upgrading nodejs version (because the old one failed to download)

This is accomplished by pinning all the plugins to a specific commit and forcing the bundler to respect the Gemfile.lock file without updating the gems.  
The specific commit for each plugin was chosen according to the [discourse-compatibility specification](https://meta.discourse.org/t/pinning-plugin-and-theme-versions-for-older-discourse-installs-discourse-compatibility/272665).

:warning: This also disables the self-hosted github runners usage that are being flaky lately.

### Checklist

- [x] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x] The documentation is generated using `src-docs`
- [x] The documentation for charmhub is updated.
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

PR #130, opened 2023-10-21
Add logging during setup process and ensure we're reacting to start hook

mthaddon:
### Overview

We need to react to the `start` event for `_set_up_discourse` rather than the `install` event, because if a container is restarted it won't get an `install` event.

Add logging to the `_set_up_discourse` method so it's easier to see at what point during the process this is failing. In production (likely due to poor disk IO related to https://bugs.launchpad.net/juju/+bug/1989004, still to be confirmed) we're seeing issues where if containers are restarted and we can end up with the set up process failing and so discourse never being started.

This only seems possible if we enter the `_set_up_discourse` method, relations are ready and we can connect to the container, but then somewhere later on in this same method we fail to complete the setup.

This should help us track down where that's happening.

### Checklist


nrobinaubertin (dismissed):

nrobinaubertin:
This is probably blocked by a dependency issue that I'm trying to solve here: https://github.com/canonical/discourse-k8s-operator/pull/125

weiiwang01 (dismissed):

arturo-seijas (dismissed):
LGTM

nrobinaubertin:
I have the same issue locally as https://github.com/canonical/discourse-k8s-operator/actions/runs/6632062737/job/18039459125#step:13:139

But not when I switch back to 'install' instead of 'start'

amandahla (approved):
LGTM

nrobinaubertin (approved):

PR #137, opened 2023-11-01
Proposal/Discussion(ISD-1274): Reduce image size

amandahla:
### Overview

Currently the ROCK results in a Docker image size of 1.49GB.  This might bring many disadvantages such as:
- Resource consumption
- Scaling
- Increased vulnerability
- Image version control

This PR intends to show an option to reduce the image size from 1.49GB to 853MB by moving the installation to the charm and by creating a storage to /var/lib/gems

TODO: 
- Discuss how handle events to avoid re-installation.
- If it's worth it considering the time to install (local test ~2min)

### Rationale

Reduce resource consumption/cost and improve scalability.

### Juju Events Changes

- on_install - where the discourse will be installed
- on_pebble_ready/config_changed - to check if the gems are installed

### Module Changes

N/A

### Library Changes

N/A

### Checklist


src/charm.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, *args):
             redis_relation={},
         )
         self._require_nginx_route()
+        self._gems_installed = False
```
mthaddon, line 103: This variable is going to be set to `False` every time the charm class is initialised, i.e. on every independent hook invocation, so I think that means this will get run on every config-changed hook?

src/charm.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, *args):
...
+        self.framework.observe(self.on.install, self._on_install)
```
mthaddon, line 104: One thing to note here is that this event won't be called if the workload container is restarted without the charm container being restarted. The only hook that's executed in that case is the discourse_pebble_ready hook. This would likely work okay because you're calling the `_on_install` hook as part of `_config_changed` (which is called by the discourse_pebble_ready event handler), but it makes it feel like we should have a layer of abstraction for the thing we're doing (installing gems) and call that from the relevant event rather than calling `self._on_install(event)` (but seem the comment above about where this should be called from).

mthaddon (commented):
This is a really interesting proposal. It feels like we may need some discussions about the trade offs.

src/charm.py
amandahla: My idea was to add only the install event but then I noticed that pebble ready event sometimes is fired first 🤔
```Note that although juju guarantees to fire install before pebble-ready, it is possible that the container is already ready by the time install fires, and can_connect would return True.```
https://juju.is/docs/sdk/container-name-pebble-ready-event

So that's why I added it to the config_changed method. But ideally, this shouldn't be needed...

Anyway, I added this variable just to see if it works, I believe it would be better/safer to check for the directory content.

File: src/charm.py (hunk 1 above)
Comment (amandahla) [reply src/charm.py]: I removed it and now the config-chaged uses a method to check if there is something on the /var/lib/gems to decide if will install them or not. I'm considering removing this check from the on_install event and assuming it will be installed/reinstalled. What do you think?

File: discourse_rock/rockcraft.yaml [hunk 3]
```diff
@@ -184,36 +185,6 @@ parts:
...
-      # The following is a fix for https://github.com/discourse/discourse-prometheus/blob/72fff206ba18ad5ca3112fed2f5f0ce6a17ca6f8/plugin.rb#L13C1-L13C105
-      sed -i "s/gem 'prometheus_exporter', .*/gem 'prometheus_exporter', '0.5.0'/g" Gemfile
-      bin/bundle install
-      bin/bundle install --gemfile="plugins/discourse-prometheus/Gemfile"
-      bin/bundle install --gemfile="plugins/discourse-saml/Gemfile"
-    organize:
-      '/var/lib/gems': var/lib/gems
-      'srv/discourse/app/bin/*': srv/discourse/app/bin/
-      'srv/discourse/app/Rakefile': srv/discourse/app/
-      'srv/discourse/app/bundle': srv/discourse/app/
```
jdkandersson, line 216: @gregory-schiano an interesting PR to look at

I'm curios what gems are used for in this case? Is it an integral part of creating an image for discourse?

A guiding principle for rocks is that they should work standalone, not just when they are used with a charm. So when someone makes use of the image generated by the rockfile, it should do something sensible. So if the gems are required for discourse then they really should be part of the rock.

It would be interesting to see what can be done to reduce the size of the image, e.g., using chisel

amandahla: I see your point and it makes sense but considering that this ROCK is being created for the charm and not to provide a Discourse instance, I don't think there is a problem in transferring the install steps to the charm itself since there is even a [Juju event](https://juju.is/docs/sdk/install-event) for that and the benefits are worthy.

src/charm.py
```diff
@@ -119,6 +121,111 @@ def __init__(self, *args):
...
+        container = self.unit.get_container(SERVICE_NAME)
+        if not container.can_connect():
+            logger.info("Not ready to install")
+            return
+        if self.model.unit.is_leader():
+            self.model.unit.status = MaintenanceStatus("Installing discourse")
+            try:
+                commands = [
+                    (False, ["rm", "-rf", "/var/lib/gems/*"]),
+                    (False, ["gem", "install", "-n", "bin", "bundler"]),
```
weiiwang01, line 135: I think gem install requires network. In that case, should the HTTP proxy be set up for it?

amandahla: Good catch! I moved it back to the ROCK

discourse_rock/rockcraft.yaml
nrobinaubertin: I'm a bit torn on this: I feel like the workload OCI image doesn't need to be something that is standalone and at the same time I feel like we shouldn't move too much of the ROCKS logic into the charm (but where's the line ? Is this PR already going too far ?)

amandahla:
Thanks for all the reviews. I'm closing this PR to no conflict with current work and to discuss the workload/charm responsibilities in a different context.

discourse_rock/rockcraft.yaml
jdkandersson: Apologies I didn't get back to this, it got lost a bit in time off and catching up. Whenever we create rocks the guiding principle is that they should be useful on their own. I know we create lots of rocks mostly for charms - eventually these will be published as sources. So similar to snaps etc, whenever we create these kind of artefacts they should not be charm specific

PR #141, opened 2023-11-17
feat: Discourse v3 rock

merkata:
This PR builds a Discourse (3.1.3) rock with updated core dependencies (ruby version, yarn and node) and adjusted patches. Not in scope is updating the plugins for Discourse and adding new ones, these will be tackled in future PRs.

The PR adds ruby 3.2.2 as Discourse v3 requires ruby > 3.2.0 and adjusts the bundle to be in accordance to https://github.com/discourse/discourse_docker/blob/main/image/base/release.Dockerfile which is the production Dockerfile for a dockerized Discourse.

Additional permissions are set under the _daemon_ home dir, which is /var/lib/pebble/default, for ruby and yarn related tasks. I did not set the main directory ownership to _daemon_ to not affect pebble in some way.


discourse_rock/patches/lp1903695.patch
```diff
@@ -1,11 +1,13 @@
...
 --- a/lib/middleware/anonymous_cache.rb
 +++ b/lib/middleware/anonymous_cache.rb
-@@ -318,7 +318,7 @@ module Middleware
+@@ -347,7 +347,7 @@ module Middleware
+       return @app.call(env) if defined?(@@disabled) && @@disabled
  
-     def call(env)
        if PAYLOAD_INVALID_REQUEST_METHODS.include?(env[Rack::REQUEST_METHOD]) &&
--        env[Rack::RACK_INPUT].size > 0
-+        env[Rack::RACK_INPUT].respond_to?(:size) && env[Rack::RACK_INPUT].size > 0
```
arturo-seijas, line 8: We probably don't need this anymore

discourse_rock/rockcraft.yaml
```diff
@@ -4,46 +4,24 @@
...
-    # Terser is needed to compile discourse's assets
-    # We don't use the system package to avoid pulling ubuntu's nodejs package
-    # Related to: https://git.launchpad.net/~weii-wang/charm-k8s-discourse/commit/?h=fix-terser-precompile&id=ada4ba97b28ab53dd0a619b3a9325b7a981afac1
-    plugin: nil
-    source: .
-    after:
-      - install-node
-    overlay-script: |
-      $CRAFT_OVERLAY/usr/bin/node $CRAFT_OVERLAY/usr/bin/npm --prefix $CRAFT_OVERLAY/usr/local/ i -g terser
+      curl -sL https://deb.nodesource.com/setup_18.x | bash -
```
nrobinaubertin, line 20: We're loosing a lot of control here: we can't specify the exact version of node nor the one of yarn, we can't verify the shasum of the binaries.
That said, I understand that the previous way was not "source first". Should we change that to rebuild node in a future version of the rockfile ?

discourse_rock/patches/lp1903695.patch
merkata: Might be, I would like to tackle that in another story though. Looking at the bug on Launchpad I don't understand why/how the bug creeps in, whether a bug/issue upstream is necessary, so I really want to dig into this once there is a running rock with Discourse v3.

discourse_rock/rockcraft.yaml
merkata: That's a valid concern, however - you should be able to specify the version of the packages in apt and in npm, it is just that we're not doing it currently. I think we should aim for apt/snap first and then source, but also I'm in favor of following the install guidelines of the upstream projects as close as possible.

discourse_rock/patches/lp1903695.patch
merkata: @arturo-seijas the Discourse code still lacks that piece of code, so I have opted for keeping it FYI.

discourse_rock/rockcraft.yaml
merkata: @nrobinaubertin the refactored version now installs the tooling in a versioned manner like the original did, so that should be fine now imo. wdyt?

discourse_rock/rockcraft.yaml
```diff
@@ -151,84 +143,70 @@ parts:
...
   apply-patches:
     plugin: nil
     build-packages:
       - git
-    after:
-      - get-discourse
-      - get-patches
+    after: [discourse, patches]
     override-stage: |
+      craftctl default
```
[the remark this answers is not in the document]
merkata: Thinking about it I don't think we need a `craftctl default` here, since the `plugin` is _nil_, so there is like...no default? So I removed it, see how this behaves now.

discourse_rock/rockcraft.yaml
```diff
@@ -151,84 +143,70 @@ parts:
...
       - libssl-dev
-    override-build: |
-      cd $CRAFT_STAGE/srv/discourse/app
-      gem install -n "bin" bundler
-      # rexml version is forced to avoid conflicting transient dependency issue
-      sed -i 's/rexml (3.2.5)/rexml (3.2.6)/' Gemfile.lock
-      # The following is a fix for https://github.com/discourse/discourse-prometheus/blob/72fff206ba18ad5ca3112fed2f5f0ce6a17ca6f8/plugin.rb#L13C1-L13C105
-      sed -i "s/gem 'prometheus_exporter', .*/gem 'prometheus_exporter', '0.5.0'/g" Gemfile
+    override-prime: |
+      craftctl default
```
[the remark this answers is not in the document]
merkata: Trve, removed. :crossed_fingers:

jdkandersson (commented):
Nice naming for the parts! 🙂

jdkandersson (approved):

yanksyoon (approved):
LGTM! I think this repository could benefit from a PR template :D

PR #146, opened 2023-11-22
Rework handling of the pebble_ready event

nrobinaubertin:
### Overview

We discovered that in cases where the workload container gets restarted on its own, only a pebble_ready event would be thrown by juju and the container would not be correctly re-planned.  
This PR handles pebble_ready event correctly and adds a named function for each event we listen to. This is a step towards "managing charm complexity", allows us to have more fine grained logging and handling of events.  
During the creation of the PR, it was noted that the redis_relation_changed handler didn't try to setup discourse as needed, so it was added.

### Checklist


amandahla:
Out of curiosity: From the doc [Lifecycle events](https://juju.is/docs/sdk/events#heading--complete-list-of-lifecycle-events), while deploying the charm, this happens: install -> config-changed -> start -> <container>-pebble-ready

Would this make the set up discourse run twice?

nrobinaubertin:
> Out of curiosity: From the doc [Lifecycle events](https://juju.is/docs/sdk/events#heading--complete-list-of-lifecycle-events), while deploying the charm, this happens: install -> config-changed -> start -> -pebble-ready
> 
> Would this make the set up discourse run twice?

Yes, indeed. This is probably why it is subject to timeouts here. I need to fix this

src/charm.py
```diff
@@ -100,11 +101,11 @@ def __init__(self, *args):
...
         self._require_nginx_route()
-        self.framework.observe(self.on.start, self._set_up_discourse)
-        self.framework.observe(self.on.upgrade_charm, self._set_up_discourse)
-        self.framework.observe(self.on.discourse_pebble_ready, self._config_changed)
-        self.framework.observe(self.on.config_changed, self._config_changed)
 
+        self.framework.observe(self.on.start, self._on_start)
+        self.framework.observe(self.on.upgrade_charm, self._on_upgrade_charm)
+        self.framework.observe(self.on.discourse_pebble_ready, self._on_discourse_pebble_ready)
+        self.framework.observe(self.on.config_changed, self._config_changed)
```
jdkandersson, line 108: Should the function be called `self._on_config_changed`?

src/charm.py
```diff
@@ -115,10 +116,44 @@ def __init__(self, *args):
...
+
+    def _on_discourse_pebble_ready(self, evt: ops.PebbleReadyEvent) -> None:
+        """Handle discourse pebble ready event.
+
+        Args:
+            event: Event triggering the discourse pebble ready event handler.
+        """
+        if not self._is_setup_completed():
+            self._set_up_discourse()
+        self._config_changed(evt)
```
jdkandersson, line 153: Generally speaking, the event handler functions should not call each other, instead the common logic of handling an event should be in a separate function which is used by both event handlers. That way it is easier to reason about how each event is handled and what logic is shared between events. For example, a developer may not expect changes to the `config-changed` event handler to have an impact on `pebble-ready`. Though it would be clear that if they changed shared logic for handling both events it would impact the handling of both events. Any thoughts on that?

tests/unit/test_charm.py
```diff
@@ -429,6 +433,20 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
+    arrange: given a deployed discourse charm with all the required relations
+    act: trigger the pebble ready event on a leader unit
+    assert: the pebble plan gets updated
+    """
+    harness = helpers.start_harness()
+
+    plan_before_event = harness.get_container_pebble_plan(CONTAINER_NAME)
+    harness.container_pebble_ready(CONTAINER_NAME)
+    plan_after_event = harness.get_container_pebble_plan(CONTAINER_NAME)
+    assert plan_before_event.__dict__ != plan_after_event.__dict__
```
jdkandersson, line 447: This just checks that the dictionary has changed, do we know that it has changed in the way we expect? E.g., the service was modified as expected?

tests/unit/test_charm.py
jdkandersson, line 1: Should there also be a change/ addition to the tests to check for the `redis_relation_changed` event handling as described in the summary?

jdkandersson (commented):
Thanks for taking a step towards moving discourse to be compliant with the architecture guidance

merkata (dismissed):
Looks good, the comments are worth having a look at.

src/charm.py
nrobinaubertin: Indeed, changed.

src/charm.py
nrobinaubertin: I've regrouped those blocs in one shared function

tests/unit/test_charm.py
nrobinaubertin: I've opted to add two asserts: one to test that `_services` is in the `__dict__` and one to assert that `discourse` is in `__dict__["_services"]`. I don't think that going further and checking all the properties is meaningful. Do you agree ?

amandahla:
While reading [this](https://discourse.charmhub.io/t/charm-lifecycle/5938#heading--notes-on-the-setup-phase) I was thinking: "why not handle only pebble ready event?"

nrobinaubertin:
> While reading [this](https://discourse.charmhub.io/t/charm-lifecycle/5938#heading--notes-on-the-setup-phase) I was thinking: "why not handle only pebble ready event?"

The pebble ready event can occur when the relations don't exist or are not ready. So if we don't defer, it doesn't allow us to be sure that the charm will go through the setup

src/charm.py
```diff
@@ -115,31 +116,61 @@ def __init__(self, *args):
...
+        """
+        self._setup_and_activate()
+
     def _on_database_created(self, _: DatabaseCreatedEvent) -> None:
         """Handle database created.
 
         Args:
             event: Event triggering the database created handler.
         """
-        if self.unit.is_leader():
```
yanksyoon, line 128: Out of curiosity, may I ask why the unit no longer needs to be the leader?

src/charm.py
```diff
@@ -150,6 +181,21 @@ def _on_database_relation_broken(self, _: RelationBrokenEvent) -> None:
...
 
+    def _on_config_changed(self, _: HookEvent) -> None:
+        """Handle config change.
+
+        Args:
+            event: Event triggering the config change handler.
+        """
+        self._change_config()
+
+    def _setup_and_activate(self) -> None:
```
yanksyoon, line 192: I'm wondering if it would help to have a docstring here :D

src/charm.py
```diff
@@ -539,43 +584,42 @@ def _run_s3_migration(self) -> None:
...
         """Configure pod using pebble and layer generated from config.
 
         Args:
             event: Event triggering the handler.
         """
-        container = self.unit.get_container(SERVICE_NAME)
+        container = self.unit.get_container(CONTAINER_NAME)
         if not self._are_relations_ready() or not container.can_connect():
-            logger.info("Not ready to do config changed action.")
+            logger.info("Not ready to do config changed action")
```
yanksyoon, line 622: I'm wondering if the control flow (checking container ready, logging) should be on the 1-to-1 mapped hook and the internal functions should only perform a single action.

yanksyoon (dismissed):
LGTM, minor comments and questions. Thank you!

tests/unit/test_charm.py
jdkandersson: Looks good, thanks

jdkandersson (dismissed):

src/charm.py
```diff
@@ -539,43 +584,42 @@ def _run_s3_migration(self) -> None:
...
+            logger.info("Discourse setup: about to set workload version")
             self._set_workload_version()
-            logger.info("Workload version set.")
+            logger.info("Discourse setup: completed")
         except ExecError as cmd_err:
             logger.exception("Setting up discourse failed with code %d.", cmd_err.exit_code)
             raise
 
-    def _config_changed(self, _: HookEvent) -> None:
+    def _change_config(self) -> None:
```
mthaddon, line 614: This seems quite a confusing function name to me - it reads as if it's changing configuration. But the docstring suggest something else. How about something like `def _configure_pod`?

src/charm.py
nrobinaubertin: It is actually checked down the line just before executing the migrations

src/charm.py
nrobinaubertin: I agree that I should [move the ifs up](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html) but I wanted to do that globally in the charm in another PR

src/charm.py
nrobinaubertin: I changed it to _configure_pod

src/charm.py
nrobinaubertin: I re-added the missing check in execute_migrations

weiiwang01 (approved):

mthaddon (approved):

PR #159, opened 2023-12-13
fix(perms): Run unicorn as non-root user

merkata:
### Overview

Runs unicorn process as non-root.

### Rationale

It is not expected that the process runs as root, it operates under /srv/discourse/app that is not root owned and operations with .git folder are failing (error: "git detected dubious ownership in repository").

### Juju Events Changes

n/a

### Module Changes

Pebble service has the `_daemon_` user added to the command.

### Library Changes

n/a

### Checklist

- [x ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x ] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x ] The documentation is generated using `src-docs`
- [x ] The documentation for charmhub is updated.
- [x ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -439,7 +439,8 @@ class DiscourseCharm(CharmBase):
                 SERVICE_NAME: {
                     "override": "replace",
                     "summary": "Discourse web application",
-                    "command": f"sh -u _daemon_ -c '{SCRIPT_PATH}/app_launch.sh'",
+                    "command": f"{SCRIPT_PATH}/app_launch.sh",
+                    "user": "_daemon_",
                     "startup": "enabled",
                     "environment": self._create_discourse_environment_settings(),
                 }
--- a/src/charm.py
+++ b/src/charm.py
@@ -439,7 +439,7 @@ class DiscourseCharm(CharmBase):
                 SERVICE_NAME: {
                     "override": "replace",
                     "summary": "Discourse web application",
-                    "command": f"sh -c '{SCRIPT_PATH}/app_launch.sh'",
+                    "command": f"sh -u _daemon_ -c '{SCRIPT_PATH}/app_launch.sh'",
                     "startup": "enabled",
                     "environment": self._create_discourse_environment_settings(),
                 }


arturo-seijas (dismissed):

cbartz (dismissed):

src/charm.py
```diff
@@ -439,7 +439,7 @@ def _create_layer_config(self) -> ops.pebble.LayerDict:
                 SERVICE_NAME: {
                     "override": "replace",
                     "summary": "Discourse web application",
-                    "command": f"sh -c '{SCRIPT_PATH}/app_launch.sh'",
+                    "command": f"sh -u _daemon_ -c '{SCRIPT_PATH}/app_launch.sh'",
```
nrobinaubertin, line 442: I propose that we launch the script directly (as root) and that we change the user inside it, while using exec. This way, the parent process will be unicorn and not the shell

merkata: That would be something like `su _daemon_ -c exec...` I guess? Not sure if that would work though, `exec` being a shell built-in and `_daemon_` not having a shell by default.

That's what we have:

```shell
root       739     1  0 Dec10 ?        00:00:00 sh -c /srv/scripts/app_launch.sh
root       740   739  0 Dec10 ?        00:00:49 unicorn master -c config/unicorn.conf.rb
root       780   740  0 Dec10 ?        00:13:03 discourse prometheus-collector
root       781   740  2 Dec10 ?        01:44:25 discourse prometheus-global-reporter
root       799   740  1 Dec10 ?        00:43:53 unicorn worker[0] -c config/unicorn.conf.rb
root       815   740  0 Dec10 ?        00:15:01 unicorn worker[1] -c config/unicorn.conf.rb
root       826   740  0 Dec10 ?        00:13:01 unicorn worker[2] -c config/unicorn.conf.rb
```

So you want to have the `sh -c` running as `root` and the child to run as `_daemon_`?

nrobinaubertin: I meant to instruct pebble to use the `_daemon_` user to launch the script. Since the script has a shebang in place, it will get executed using `/bin/bash` and the last line of the script calls the built-in `exec` to replace the parent process

jdkandersson (dismissed):

cbartz (approved):

arturo-seijas (approved):

PR #164, opened 2024-01-08
feat(replies): Reply via pop3 polling

merkata:
Note: this PR works in theory and does not currently ship with an integration test - reason being to test this, you need two mail servers (receiving mail into Discourse, configured server with pop3 enabled, sending mails to Discourse) and a mechanism to respond to a thread via e-mail and ensure it is received. Overall a painful and tricky business. Proposal is to test this manually.

Second note: if already configured via GUI, the changes can/will get overwritten when moving to env variables, that is by design, see [here](https://meta.discourse.org/t/s3-backup-bucket-option-missing-from-settings-backups/224802).

### Overview

This PR adds the capability to enable replies to threads via pop3 polling as described [here](https://meta.discourse.org/t/set-up-reply-by-email-with-pop3-polling/14003) via `juju config` on the parameters.

### Rationale

This capability is available via GUI and it was requested that it is available as a config as well.

### Juju Events Changes

none

### Module Changes

Adds new pop3 parameters to `config.yaml` and sets these as env variables to feed to pebble in `src/charm.py`.

### Library Changes

none

### Checklist

- [x ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x ] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x ] The documentation is generated using `src-docs`
- [x ] The documentation for charmhub is updated.
- [x ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -134,3 +134,27 @@ options:
     type: boolean
     description: Configure Discourse to use https.
     default: false
+  reply_by_email_address:
+    type: string
+    description: Replies email in the form of replies+%{reply_key}@example.com.
+    default: ""
+  pop3_polling_username:
+    type: string
+    description: Username of the email account like replies@example.com.
+    default: ""
+  pop3_polling_password:
+    type: string
+    description: The password of that email account.
+    default: ""
+  pop3_polling_host:
+    type: string
+    description: Polling host such as pop.gmail.com.
+    default: ""
+  pop3_polling_enabled:
+    type: boolean
+    description: Whether pop3 polling is enabled.
+    default: false
+  reply_by_email_enabled:
+    type: boolean
+    description: Whether replying by email is enabled.
+    default: false
--- a/src/charm.py
+++ b/src/charm.py
@@ -401,6 +401,12 @@ class DiscourseCharm(CharmBase):
             "DISCOURSE_SMTP_PASSWORD": self.config["smtp_password"],
             "DISCOURSE_SMTP_PORT": str(self.config["smtp_port"]),
             "DISCOURSE_SMTP_USER_NAME": self.config["smtp_username"],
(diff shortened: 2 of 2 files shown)


jdkandersson (approved):

config.yaml
```diff
@@ -134,3 +134,27 @@ options:
...
+    default: ""
+  pop3_polling_password:
+    type: string
+    description: The password of that email account.
+    default: ""
+  pop3_polling_host:
+    type: string
+    description: Polling host such as pop.gmail.com.
+    default: ""
+  pop3_polling_enabled:
```
cbartz, line 153: What happens if this flag is enabled but the other configuration options are not provided? Shouldn't there be some sort of non-emptiness check? What does the GUI do in this case?

nrobinaubertin: Passing some environment variables has the same effect as using the GUI, so discourse will not enable the polling in that case but will not complain (we may see some alert in the admin settings)

nrobinaubertin: I agree that we should probably add some logic to put a warning in the status of the charm, or an error

amandahla (approved):

config.yaml
merkata: We might think about this, though like Niels said - Discourse should handle that. The best we could do is to put a warning maybe, as you could configure these settings in any order and once you have them, polling will work. If not, we are guiding the user to set them all at once, like `juju config a=x b=y..` on one line - which is also an option I guess, then we need to update the documentation to reflect that this is how we expect the charm to be configured in terms of polling.

merkata:
Original RT is closed, so I’m closing this one as well and it can be reopened if needed.

PR #171, opened 2024-01-15
Add rolling-ops library

nrobinaubertin:
### Overview

Currently there is no locking during charm upgrade. This means that we rely entirely on juju for ordering of unit upgrades. This can mean that during an upgrade where you have two units the second unit’s upgrade can begin before the first unit is ready to accept traffic (it may be compiling assets or running s3 migrations as part of the config-changed hook that comes after a charm upgrade hook).  

This PR adds the rolling-ops library to handle our charm upgrades.

### Juju Events Changes

The `upgrade-charm` event fires now an `acquire_lock` that will be used by the `rollingèops` library to roll out changes gradually.

### Checklist

### DIFF
--- a/lib/charms/rolling_ops/v0/rollingops.py
+++ b/lib/charms/rolling_ops/v0/rollingops.py
@@ -1,0 +1,414 @@
+# Copyright 2022 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This library enables "rolling" operations across units of a charmed Application.
+
+For example, a charm author might use this library to implement a "rolling restart", in
+which all units in an application restart their workload, but no two units execute the
+restart at the same time.
+
+To implement the rolling restart, a charm author would do the following:
+
+1. Add a peer relation called 'restart' to a charm's `metadata.yaml`:
+```yaml
+peers:
+    restart:
+        interface: rolling_op
+```
+
+Import this library into src/charm.py, and initialize a RollingOpsManager in the Charm's
+`__init__`. The Charm should also define a callback routine, which will be executed when
+a unit holds the distributed lock:
+
+src/charm.py
+```python
+# ...
+from charms.rolling_ops.v0.rollingops import RollingOpsManager
+# ...
+class SomeCharm(...):
+    def __init__(...)
+        # ...
+        self.restart_manager = RollingOpsManager(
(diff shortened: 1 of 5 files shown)


gtrkiller (approved):
LGTM

jdkandersson (approved):

jdkandersson:
Perhaps it is worthwhile to create an integration test that demonstrates the rolling restarts? Just to make sure we're using the library correctly?

merkata (approved):
:partying_face: :tada: lgtm

merkata:
> Perhaps it is worthwhile to create an integration test that demonstrates the rolling restarts? Just to make sure we're using the library correctly?

Yes, we did discuss this with @nrobinaubertin and @mthaddon - initially the idea is to have unit tests (they cover imo that rolling ops is used as intended and should make sure we don't introduce regressions) and think about what integration test to have to cover the aspect of rolling ops as well. 

Looking at the integration tests I have an idea - we might want to introduce scaling the app (I think we test with a single unit), then think of an upgrade test - this is akin to what [postgresql charm](https://github.com/canonical/postgresql-operator/blob/main/tests/integration/ha_tests/test_upgrade.py) is testing, they don't explicitly test rolling ops afaik. A basic smoke test should be also fine, similar to the [example](https://github.com/canonical/charm-rolling-ops/blob/main/tests/integration/smoke_test.py) in rolling ops lib - this requires an action for a restart in place if we do it in the same manner. 

Needs to be further discussed and refined, though we're definitely discussing it. The upgrade itself doesn't necessarily come with downtime and I'm not sure we want to have a test that can always cause downtime and we prove it to be fixed via locking through rolling ops, might be too much. Anyway, these are solely my thoughts. What do you think about this @jdkandersson ? Do you have also suggestions on the integration tests? Thanks for the comment!

jdkandersson:
> > Perhaps it is worthwhile to create an integration test that demonstrates the rolling restarts? Just to make sure we're using the library correctly?
> 
> Yes, we did discuss this with @nrobinaubertin and @mthaddon - initially the idea is to have unit tests (they cover imo that rolling ops is used as intended and should make sure we don't introduce regressions) and think about what integration test to have to cover the aspect of rolling ops as well.
> 
> Looking at the integration tests I have an idea - we might want to introduce scaling the app (I think we test with a single unit), then think of an upgrade test - this is akin to what [postgresql charm](https://github.com/canonical/postgresql-operator/blob/main/tests/integration/ha_tests/test_upgrade.py) is testing, they don't explicitly test rolling ops afaik. A basic smoke test should be also fine, similar to the [example](https://github.com/canonical/charm-rolling-ops/blob/main/tests/integration/smoke_test.py) in rolling ops lib - this requires an action for a restart in place if we do it in the same manner.
> 
> Needs to be further discussed and refined, though we're definitely discussing it. The upgrade itself doesn't necessarily come with downtime and I'm not sure we want to have a test that can always cause downtime and we prove it to be fixed via locking through rolling ops, might be too much. Anyway, these are solely my thoughts. What do you think about this @jdkandersson ? Do you have also suggestions on the integration tests? Thanks for the comment!

@merkata that sounds good. I think what the rolling ops does is that it ensures that the upgrades are done one unit at a time? Perhaps the integration test could trigger and upgrade and just make sure that the app is available throughout the upgrade?

PR #178, opened 2024-01-17
[Do not merge] Update charmcraft builder from 20.04 to 22.04

nrobinaubertin:
### Overview

Update charmcraft builder from 20.04 to 22.04

### Checklist

- [x] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x] The documentation is generated using `src-docs`
- [x] The documentation for charmhub is updated.
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

PR #199, opened 2024-03-06
Add SAML integration

gtrkiller:
### Overview

Add SAML integration using the SAML Integrator charm

### Rationale

Simplify the charm's logic and modularize SAML interactions.

### Library Changes

Add the SAML

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The documentation is generated using `src-docs`
- [X] The documentation for charmhub is updated.
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


charmcraft.yaml
```diff
@@ -6,10 +6,10 @@ type: "charm"
 bases:
   - build-on:
     - name: "ubuntu"
-      channel: "20.04"
+      channel: "22.04"
     run-on:
     - name: "ubuntu"
-      channel: "20.04"
```
nrobinaubertin, line 12: Is this okay now ? (cf: https://github.com/canonical/discourse-k8s-operator/pull/178#pullrequestreview-1829558666)

tests/integration/test_charm.py
```diff
@@ -196,6 +196,9 @@ async def test_saml_login(  # pylint: disable=too-many-locals,too-many-arguments
     act: add an admin user and enable force-https mode.
     assert: user can login discourse using SAML Authentication.
     """
+    saml_app = await model.deploy("saml-integrator", channel="edge", series="jammy", trust=True, config={"entity_id": "https://login.staging.ubuntu.com", "metadata_url": "https://login.staging.ubuntu.com/saml/metadata"})
```
nrobinaubertin, line 199: I would rather use the SamlK8sTestHelper (see here: https://github.com/canonical/synapse-operator/blob/main/tests/integration/test_nginx.py#L44) instead of relying on our infra for these tests

charmcraft.yaml
gtrkiller: I don't know if it is OK yet, but it is required for the SAML library to work (and building in 22.04 and running in 20.04 does not do the trick). Will ask around.

tests/integration/test_charm.py
weiiwang01: Hi, if you would like to use that, you might need to [add `saml-test-idp` to the integration dependencies](https://github.com/canonical/synapse-operator/blob/60b1f0bc7e754cbe39b3f32f2b17cfde25e55fc8/tox.ini#L128). There are some documents about how to use `saml-test-idp` in here https://github.com/canonical/saml-test-idp, please let me know if you have any questions 😀

charmcraft.yaml
mthaddon: Out of interest can you include some info on what's failing on 20.04? Maybe we can help get it running on that version.

gtrkiller: Sure thing! The charm fails because of type hints on the library that are not compatible with the Python version shipped with Focal e.g. https://pastebin.canonical.com/p/dSNgsQHsJQ/

mthaddon: Ah great, thanks. We own that repository/library so we can potentially propose a change to https://github.com/canonical/saml-integrator-operator to make it focal compatible if we want to go that path.

tests/integration/conftest.py
```diff
@@ -162,7 +161,7 @@ async def app_fixture(
             resources=resources,
             application_name=app_name,
             config=app_config,
-            series="focal",
+            series="jammy",
```
mthaddon, line 164: Should this revert to the original "focal" declaration?

requirements.txt
```diff
@@ -1 +1,2 @@
 ops-lib-pgsql
+pydantic==1.10.14
```
yanksyoon, line 2: Im wondering if we should pin all the way down to patch versions - should we be using poetry for this?

src/charm.py
```diff
@@ -193,6 +198,15 @@ def _on_config_changed(self, _: HookEvent) -> None:
         """
         self._configure_pod()
 
+    def _on_saml_data_available(self, event: SamlDataAvailableEvent) -> None:
+        """Handle SAML data available."""
+        fingerprint = hashlib.sha1(base64.b64decode(event.certificates[0])).hexdigest()  # nosec
```
yanksyoon, line 203: I think it would help be informative with a comment on #nosec :)

src/charm.py
```diff
@@ -193,6 +198,15 @@ def _on_config_changed(self, _: HookEvent) -> None:
...
+        relation = self.model.get_relation("saml")
+        # Will ignore union-attr since asserting the relation type will make bandit complain.
+        relation.data[self.app].update({"fingerprint": fingerprint})  # type: ignore[union-attr]
+        relation.data[self.app].update({"entity_id": event.entity_id})  # type: ignore[union-attr]
```
yanksyoon, line 207: Could these be updated with a single call?

yanksyoon (dismissed):
Minor comments! Thanks!

src/charm.py
```diff
@@ -193,6 +198,15 @@ def _on_config_changed(self, _: HookEvent) -> None:
         """
         self._configure_pod()
 
+    def _on_saml_data_available(self, event: SamlDataAvailableEvent) -> None:
+        """Handle SAML data available."""
+        fingerprint = hashlib.sha1(base64.b64decode(event.certificates[0])).hexdigest()  # nosec
+        relation = self.model.get_relation("saml")
+        # Will ignore union-attr since asserting the relation type will make bandit complain.
+        relation.data[self.app].update({"fingerprint": fingerprint})  # type: ignore[union-attr]
```
javierdelapuente, line 206: IINM, discourse can have more than one unit. If that is true, the non-leader units cannot update that part of the databag.

requirements.txt
gtrkiller: In this case we should pin down due to SAML requiring pydantic<2.

src/charm.py
javierdelapuente, line 207: Is that `entity_id` used?

src/charm.py
```diff
@@ -284,22 +298,22 @@ def _get_saml_config(self) -> typing.Dict[str, typing.Any]:
...
+            and relation.data[relation.app]
+        ):
+            saml_config["DISCOURSE_SAML_TARGET_URL"] = relation.data[relation.app][
+                "single_sign_on_service_redirect_url"
+            ]
             saml_config["DISCOURSE_SAML_FULL_SCREEN_LOGIN"] = (
                 "true" if self.config["force_saml_login"] else "false"
             )
-            fingerprint = saml_fingerprints.get(self.config["saml_target_url"])
+            fingerprint = relation.data[self.app].get("fingerprint")
```
javierdelapuente, line 316: instead of calculating the fingerprint in `_on_saml_data_available`, why not do it here (and that way it is not necessary to write anything in the databag)?

src/charm.py
gtrkiller: not anymore, thank you for spotting that!

src/charm.py
gtrkiller: Because the fingerprint is being calculated based on the certificate available in the event specifically, so I prefer to do the calculations in the event listener.

config.yaml
```diff
@@ -25,10 +25,6 @@ options:
     type: string
     description: "Comma-separated list of groups to sync from SAML provider."
     default: ""
-  saml_target_url:
```
arturo-seijas, line 28: Have you checked if this is referenced in the charm docs?

src/charm.py
```diff
@@ -193,6 +198,22 @@ def _on_config_changed(self, _: HookEvent) -> None:
         """
         self._configure_pod()
 
+    def _on_saml_data_available(self, event: SamlDataAvailableEvent) -> None:
+        """Handle SAML data available."""
+        if self.unit.is_leader():
+            # Utilizing the SHA1 hash is safe in this case, so a nosec ignore will be put in place.
+            fingerprint = hashlib.sha1(
+                base64.b64decode(event.certificates[0])
```
arturo-seijas, line 206: Are these `certificates` always non empty?

src/charm.py
```diff
@@ -193,6 +198,22 @@ def _on_config_changed(self, _: HookEvent) -> None:
...
+            ).hexdigest()  # nosec
+            relation = self.model.get_relation("saml")
```
arturo-seijas, line 208: Can you use the constant `DEFAULT_RELATION_NAME` defined in the lib instead?

src/charm.py
```diff
@@ -258,14 +279,14 @@ def _is_config_valid(self) -> bool:
         if self.config["throttle_level"] not in THROTTLE_LEVELS:
             errors.append(f"throttle_level must be one of: {' '.join(THROTTLE_LEVELS.keys())}")
 
-        if self.config["force_saml_login"] and not self.config["saml_target_url"]:
-            errors.append("force_saml_login can not be true without a saml_target_url")
+        if self.config["force_saml_login"] and self.model.get_relation("saml") is None:
```
arturo-seijas, line 282: Same here

src/charm.py
gtrkiller: Yes, they are.

tests/integration/test_charm.py
```diff
@@ -16,6 +16,7 @@
 from botocore.config import Config
 from ops.model import ActiveStatus, Application
 from pytest_operator.plugin import Model
+from saml_test_helper import SamlK8sTestHelper  # pylint: disable=import-error
```
arturo-seijas, line 19: Why doesn't pylint like this?

gtrkiller: This is the failure on the linter: tests/integration/test_charm.py:19:0: E0401: Unable to import 'saml_test_helper' (import-error)

tox.ini
```diff
@@ -79,7 +79,7 @@ commands =
 [testenv:unit]
 description = Run unit tests
 deps =
-    cosl
+    cosl==0.0.8
```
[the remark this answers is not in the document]
arturo-seijas: Oh

arturo-seijas (approved):

mthaddon (approved):
LGTM, thx!

src/charm.py
weiiwang01: I think it would a quality of life improvement if users could access SAML information (such as `entity_id`, `metadata_url`, `certificates`, and `endpoints`) outside the SAML relationship event handler (`SamlDataAvailableEvent`). What do you think? @arturo-seijas

src/charm.py
```diff
@@ -284,22 +318,22 @@ def _get_saml_config(self) -> typing.Dict[str, typing.Any]:
...
+        relation = self.model.get_relation(DEFAULT_RELATION_NAME)
+        if (
+            relation is not None
+            and relation.data[self.app]
+            and relation.app
+            and relation.data[relation.app]
+        ):
+            saml_config["DISCOURSE_SAML_TARGET_URL"] = relation.data[relation.app][
+                "single_sign_on_service_redirect_url"
+            ]
```
weiiwang01, line 332: Similar to the comment below, I think we can improve this part (accessing the relation data directly) if the SAML charm library can provide the `endpoints` information outside of the event handler.

weiiwang01 (approved):
left a few minor suggestions for future improvement, but overall looks good!

Issue #208, opened 2024-03-31
The S3 upload_assets routine can fail on pod startup, resulting in pages referencing S3-hosted files that don't exist


barryprice:
### Bug Description

![image (1)](https://github.com/canonical/discourse-k8s-operator/assets/21195412/88cf2513-ad57-4f5f-8b7d-993cb1039805)

It's possible for Discourse to fail its S3 upload_assets routine after a restart, resulting in errors like the one above.

The HTML served from the pods loads fine, but the e.g. Javascript URLs referenced within are for paths expected to be created by the upload_assets routine, but which were never uploaded to S3.

This results in a broken page with a spinner, the main content never appears.

### To Reproduce

Deploy the application with s3_enabled=True and all associated config set.

Wait for (or force) a restart.

Be unlucky enough to experience this bug (it's unclear how/why it happened).

### Environment

```
prod-discourse-ubuntu-com-k8s@is-bastion-ps6:~$ juju version
3.1.6-ubuntu-amd64
prod-discourse-ubuntu-com-k8s@is-bastion-ps6:~$ juju status
Model                          Controller    Cloud/Region         Version  SLA          Timestamp
prod-discourse-ubuntu-com-k8s  prodstack-is  k8s-prod-is/default  3.1.6    unsupported  14:59:38Z

SAAS        Status  Store         URL
grafana     active  local         admin/prod-cos-k8s-ps6-is-charms.grafana
loki        active  local         admin/prod-cos-k8s-ps6-is-charms.loki
postgresql  active  prodstack-is  admin/prod-discourse-ubuntu-com-db.postgresql
prometheus  active  local         admin/prod-cos-k8s-ps6-is-charms.prometheus

App                       Version  Status  Scale  Charm                     Channel        Rev  Address        Exposed  Message
discourse-k8s             3.2.0    active      2  discourse-k8s             latest/stable   95  10.87.199.71   no       
nginx-ingress-integrator  25.3.0   active      1  nginx-ingress-integrator  latest/stable   81  10.87.154.21   no       
redis-k8s                 7.0.4    active      1  redis-k8s                 latest/edge     26  10.87.238.179  no       

Unit                         Workload  Agent  Address          Ports  Message
discourse-k8s/0              active    idle   192.168.102.255         
discourse-k8s/1*             active    idle   192.168.103.157         
nginx-ingress-integrator/0*  active    idle   192.168.102.249         
redis-k8s/0*                 active    idle   192.168.102.248         
[...]

PR #225, opened 2024-04-19
Redis integration. Do not use StoredState.

javierdelapuente:
### Overview

Related to https://github.com/canonical/discourse-k8s-operator/issues/208 (failing S3 upload_assets). Also story ISD-1767.

When the Redis IP address changes, if discourse also restarts/upgrades, it can get into error state, and it will not advance. 
 The charm will be unable to migrate/upload to s3 or any other similar action in the workload, as the Redis IP address is wrong.  The reason for the Redis IP address to be wrong is because it was stored previously in StoredState, being the correct one is in the relation data of the Redis integration.

The fix consists in not using StoredState and getting the IP address from the relation data.

### Rationale

There is a fix that can get Discourse into error state, without any fix except changing the charm code manually (or undeploying the charm).

### Checklist

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -412,7 +412,7 @@ def test_start_when_leader():
     act: trigger the start event on a leader unit
     assert: migrations are executed and assets are precompiled.
     """
-    harness = helpers.start_harness()
+    harness = helpers.start_harness(run_initial_hooks=False)
 
     # exec calls that we want to monitor
     exec_calls = [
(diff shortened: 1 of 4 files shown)


tests/unit/test_charm.py
```diff
@@ -388,22 +388,6 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
     charm._on_anonymize_user_action(event)
 
 
-def test_handle_pebble_ready_event():
```
javierdelapuente, line 391: This test is not doing what it says. The plan before the pebble ready was empty because of Redis integration.

mthaddon (approved):

arturo-seijas (approved):

PR #228, opened 2024-04-23
ISD-1625 Add enable_plugins action

Thanhphan1147:
### Checklist


src/discourse.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Interface for interactions with discourse."""
+import logging
+import typing
+
+from ops.pebble import ChangeError, ExecError, ExecProcess
+
+CONTAINER_APP_USERNAME = "_daemon_"
```
mthaddon, line 10: I think we define this in `src/charm.py` currently, so can we either refer to that one here, or refer to this one there?

src/discourse.py
```diff
@@ -0,0 +1,71 @@
...
+DISCOURSE_PATH = "/srv/discourse/app"
```
mthaddon, line 11: Same as above, we define this in `src/charm.py` currently.

src/discourse.py
Thanhphan1147: I think the end goal is to have the constant defined here and refer to it in `charm.py`. But imo refactoring `charm.py` is quite a significant change, so I'd prefer to keep things simple in this PR, then work on the refactoring on another PR.

The idea is also to move most of the calls to `bundle exec rake ...` from `charm.py` to this module. So I think we can leave it like this for now, then if this is ok for @merkata and @nrobinaubertin, do a follow-up PR refactoring `charm.py`.

What do you think?

src/discourse.py
Thanhphan1147: I replied to the above comment.

src/discourse.py
mthaddon: Sure, if they're okay with it, sounds good to me too.

src/discourse.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, msg: str):
+        """Initialize a new instance of the CharmConfigInvalidError exception.
+
+        Args:
+            msg: Explanation of the error.
+        """
+        self.msg = msg
+
+
+def update_site_settings(
```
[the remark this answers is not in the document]
mthaddon: Ok, so we're only using for installing plugins, but it can be used for other settings? If so I think it's fine to leave as is.

src/charm.py
```diff
@@ -801,6 +803,24 @@ def _on_anonymize_user_action(self, event: ActionEvent) -> None:
...
                 )
 
+    def _on_enable_plugins_action(self, event: ActionEvent) -> None:
+        """Enable discourse plugin(s).
+
+        Args:
+            event: Event triggering the anonymize_user action.
+        """
+        plugins_comma_delimited = event.params["plugins"]
+        plugins = plugins_comma_delimited.split(",")
```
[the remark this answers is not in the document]
Thanhphan1147: The rake task can gracefully handle errors if the setting doesn't exist and simply ignores that setting and prints out a message at the end saying what was ignored.

src/discourse.py
```diff
@@ -0,0 +1,71 @@
...
+
+def enable_plugins(environment: typing.Dict[str, str], container, plugins: typing.List[str]):
+    """Enable discourse plugins."""
+    site_settings_config: typing.Dict[str, str] = {
+        f"{plugin}_enabled": "true" for plugin in plugins
+    }
+    return update_site_settings(container, environment, site_settings_config)
+
+
+def disable_plugins(environment: typing.Dict[str, str], container, plugins: typing.List[str]):
```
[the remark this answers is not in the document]
Thanhphan1147: I added this for the sake of having a method to disable the plugins if we ever need to, but indeed we don't actually use this anywhere. Should I remove it? Or I can also create another action for disabling plugins.

PR #230, opened 2024-04-24
enable plugins in integration tests

Thanhphan1147:
Enable all plugins when running integration tests using the `rake site_settings:import` task

```
echo '{settings}' | /srv/discourse/app/bin/bundle exec rake site_settings:import -
```

### Checklist

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -208,6 +208,7 @@ async def app_fixture(
         '''/srv/discourse/app/bin/bundle exec rake site_settings:import -"'''
     )
 
+    logger.info("Enabling plugins: %s", enable_plugins_command)
     action = await unit.run(f"/bin/bash -c '{enable_plugins_command}'")
     logger.info(action.results)
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -210,6 +210,7 @@ async def app_fixture(
 
     logger.info("Enabling plugins: %s", enable_plugins_command)
     action = await unit.run(f"/bin/bash -c '{enable_plugins_command}'")
+    await action.wait()
     logger.info(action.results)
 
     yield application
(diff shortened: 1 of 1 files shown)


javierdelapuente (dismissed):

merkata (dismissed):
Nice!!

tests/integration/conftest.py
```diff
@@ -193,6 +201,14 @@ async def app_fixture(
...
     )
     await model.wait_for_idle(status="active", raise_on_error=False)
+    inline_yaml = "\n".join(f"{plugin}_enabled: true" for plugin in ENABLED_PLUGINS)
+    enable_plugins_command = (
+        "pebble exec --user=_daemon_ --context=discourse -w=/srv/discourse/app -ti -- /bin/bash -c "
+        f'''"echo '{inline_yaml}' | /srv/discourse/app/bin/bundle exec rake site_settings:import -"'''
+    )
+
+    action = await unit.run(f"/bin/bash -c '{enable_plugins_command}'")
+    logger.info(action.results)
```
yanksyoon, line 211: Out of curiosity, would this action be optional and not need to be awaited?

yanksyoon (dismissed):
LGTM!

tests/integration/conftest.py
Thanhphan1147: I think we should await the completion of this action as we want to test that discourse can function properly with all of the plugins enabled.

yanksyoon: Oh sorry for the vague statement, let me try to clarify :D
```
action = await unit.run(block=True) # note the block=True here or
await action.wait() # to wait for the results
assert action.results.["property_here"] == "Expected results"
```

From the juju docs of the action:
> blocl: Useful with older versions of juju, i.e. getting the result without having to call action.wait() separately.
:returns: A juju.action.Action instance.
Note that this is very similarly to unit.run_action only enqueues the action. You will need to call action.wait() on the resulting Action instance if you wish to block until the action is complete.

yanksyoon (dismissed):

javierdelapuente (approved):

merkata (approved):

PR #239, opened 2024-04-30
Integration test for upgrades

javierdelapuente:
### Overview

The goal of this PR is to write an integration test that tests the charm upgrade process. There should be no downtime in the upgrade process, so the way of checking it in this test is to make a request every 0.5 seconds to the endpoint `/srv/status` (to the ingress endpoint), in a similar way as how it will work in a real environment.

Two issues were found in the test.

Lack of a graceful shutdown. Kubernetes (pebble) sends a SIGTERM signal to Discourse. However, `unicorn` requires SIGQUIT instead for a graceful shutdown (see https://yhbt.net/unicorn/SIGNALS.html). Besides, it looks that having some time before sending the signal could help (see https://medium.com/@acgs771126/ror-unicorn-graceful-shutdown-in-kubernetes-f66d83e1eb48). This is similar to what is suggested in Heroku (see https://blog.heroku.com/unicorn_rails). For that, a patch has been generated that translates SIGTERM to SIGQUIT, and also a delay before sending SIGQUIT to the unicorn master.

Workload containers are ready even when there is no discourse running. To improve this situation, the pebble check  `discourse-setup-completed` has been moved to the rock, so it always runs. Before that, the check was only added when the discourse service was also added.

### Checklist


mthaddon (approved):
Looks great, thanks!

merkata (approved):
:clap: nicely done!

PR #241, opened 2024-05-02
Revisited add-admin action

codethulu:
### Overview

Created an `add_user` action, which simply creates a user account. The `on_add_admin` method then is responsible for calling this method, and then promoting the user to admin level. This behaviour has effectively been split.

### Rationale

We want an action to change an existing user to an admin and an action to create a user. And if the user already exists, we don’t want to update the password. If the user doesn’t exist, we can to create the user and generate a random secure password for them. 

### Juju Events Changes

N/A

### Module Changes

N/A

### Library Changes

No significant changes, just updates to be consistent with main.

### Checklist


codethulu:
There were a few methods that needed to be considered. In the end, after discussion with @nrobinaubertin and @amandahla I have decided to create a new rake patch - 

This approach was chosen for a few reasons:
- This does not append to the end of a pre-existing file. This is a more robust solution as it is not dependent on replacing some code. Can lead to less conflicts, as all additional processes are in one file.
- Clear to see all (non-merged upstream) rake commands we have written
- Clearly defines a `user:create` and `admin:promote` endpoints. Although technically, we could use the `admin:create` endpoint and pass in different parameters, to create a user and input `n` to not grant admin privileges, this does not seem entirely clear in the code without referencing the upstream definition of the method. Additionally, both me and Amanda agree that such methods, especially the `user:create`, should exist in the user rake commands. I want to submit a PR to discourse to attempt to have these supported upstream.

codethulu:
Additionally, the previous rake patch to add `user:anonymize` has been removed, since this has been [merged to discourse](https://github.com/discourse/discourse/blob/v3.2.0/lib/tasks/users.rake#L187) in version 3.2.0, which is what we use in rockcraft.

mthaddon (commented):
Haven't done a full review, but are we meaning to remove the `anonymize_user.patch` file here? Having a full description of the change and the rationale would be good to confirm what we're intending to do and why.

nrobinaubertin:
> rather places additional rake commands in a `canonical.rake` file

I would name it something like `discourse-charm.rake` to avoid tying it directly to the Canonical name (it's not a patch specifically for our needs but for the charm's needs)

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
...
++  if !email || email !~ /@/
++    puts "ERROR: Expecting rake admin:promote[some@email.com]"
++    exit 1
++  end
++  if User.find_by_email(email).nil?
++    puts "ERROR: User with email #{email} not found"
++    exit 1
++  end
++ admin = User.find_by_email(email)
++ admin.grant_admin!
```
arturo-seijas, line 18: What's this exclamation point at the end?

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
...
++ admin.activate
++ say("\nYour account now has Admin privileges!")
++end
++
++desc "Create the user for given credentials"
++task "users:create" => :environment do
++  require "highline/import"
++
++  begin
++    email = ask("Email:  ")
```
arturo-seijas, line 30: Why prompting for the user instead of passing it as a cmd line arg?

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
...
++    existing_user = User.find_by_email(email)
++
++    # check if user account already exists
++    if existing_user
++      # user already exists, ask for password reset
++      new_user = existing_user
++      reset_password =
++        ask(
++          "User with this email already exists! Do you want to reset the password for this email? (Y/n)  ",
```
arturo-seijas, line 39: I would return an error and add a flag to reset password if exists

discourse_rock/patches/rake.patch
codethulu: upstream (discourse) method

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
...
++# frozen_string_literal: true
++
++desc "Promote a user to admin"
++task "admin:promote", [:email] => [:environment] do |_, args|
++email = args[:email]
++  if !email || email !~ /@/
++    puts "ERROR: Expecting rake admin:promote[some@email.com]"
++    exit 1
++  end
++  if User.find_by_email(email).nil?
```
javierdelapuente, line 13: we are calling `find_by_email` twice. Either put  `admin = User.find_by_email(email)` before the `if` and check for `nil` or do it all together in the if (assign and check not nil).

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
...
++    puts "ERROR: User with email #{email} not found"
++    exit 1
++  end
++ admin = User.find_by_email(email)
++ admin.grant_admin!
++ admin.change_trust_level!(1) if admin.trust_level < 1
++ admin.email_tokens.update_all confirmed: true
```
javierdelapuente, line 20: Is this necessary? Are we creating any new email token?

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
...
++ admin.activate
```
javierdelapuente, line 21: Wasn't the user activated?

src/charm.py
```diff
@@ -701,33 +701,63 @@ def _on_add_admin_user_action(self, event: ActionEvent) -> None:
...
+        password = event.params["password"]
+
+        process = container.exec(
+            [
+                os.path.join(DISCOURSE_PATH, "bin/bundle"),
+                "exec",
+                "rake",
+                "users:create",
+            ],
+            stdin=f"{email}\n{password}\n",
```
javierdelapuente, line 751: Having a quick look at the ruby code, you require the password twice.

src/charm.py
```diff
@@ -701,33 +701,63 @@ def _on_add_admin_user_action(self, event: ActionEvent) -> None:
...
-                user=CONTAINER_APP_USERNAME,
-                environment=self._create_discourse_environment_settings(),
-                timeout=60,
+
+        email = event.params["email"]
+
+        if not container.can_connect():
+            event.fail("Unable to connect to container, container is not ready")
+            return
+        self._create_user(event)
```
javierdelapuente, line 711: If the create_user fails, it continues running this code. Does not sound like correct.

src/charm.py
```diff
@@ -701,33 +701,63 @@ def _on_add_admin_user_action(self, event: ActionEvent) -> None:
...
+                "admin:promote",
+                email,
+            ],
+            working_dir=DISCOURSE_PATH,
+            user=CONTAINER_APP_USERNAME,
+            environment=self._create_discourse_environment_settings(),
+        )
+        try:
+            process.wait_output()
+            event.set_results(email)
```
javierdelapuente, line 727: set_results should be a Dicts. Any reason to change the previous results?

src/charm.py
```diff
@@ -701,33 +701,63 @@ def _on_add_admin_user_action(self, event: ActionEvent) -> None:
...
-            except ExecError as ex:
-                event.fail(
-                    # Parameter validation errors are printed to stdout
-                    f"Failed to create user with email {email}: {ex.stdout}"  # type: ignore
-                )
-        else:
-            event.fail("Container is not ready")
+
+    def _create_user(self, event: ActionEvent):
+        """Create a new user in Discourse.
```
javierdelapuente, line 734: Maybe this function could receive an email and a password and not an event, and so be more generic. The only use of the event is to set the action to failed, which maybe it could be in the calling function.

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
+--- a/lib/tasks/discourse-charm.rake
++++ b/lib/tasks/discourse-charm.rake
+@@ -0,0 +1,84 @@
++# frozen_string_literal: true
++
++desc "Promote a user to admin"
++task "admin:promote", [:email] => [:environment] do |_, args|
```
javierdelapuente, line 7: We are creating code that cannot be tested by the charm unit tests. We are testing by luck in the integration tests. I would rather create tests in Ruby, or at least have one specific integration test in the charm that tests this code on purpose.

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
+--- a/lib/tasks/discourse-charm.rake
++++ b/lib/tasks/discourse-charm.rake
```
javierdelapuente, line 2: I would change the name of the file from `rake.patch` to something more explicit, like `discourse-charm.rake` or `user-tasks.rake`

discourse_rock/patches/rake.patch
```diff
@@ -0,0 +1,87 @@
...
++  end
++ admin = User.find_by_email(email)
++ admin.grant_admin!
++ admin.change_trust_level!(1) if admin.trust_level < 1
++ admin.email_tokens.update_all confirmed: true
++ admin.activate
++ say("\nYour account now has Admin privileges!")
++end
++
++desc "Create the user for given credentials"
```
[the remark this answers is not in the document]
codethulu: https://github.com/canonical/discourse-k8s-operator/pull/241#issuecomment-2112981001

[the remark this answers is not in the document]
codethulu: tried to explain some reasoning here

src/charm.py
codethulu: this is inteded behaviour from the original method (https://github.com/canonical/discourse-k8s-operator/blob/main/src/charm.py#L723)

javierdelapuente: `event.set_results({"user": f"{email}"})`   !=   `event.set_results(email)`

discourse_rock/patches/rake.patch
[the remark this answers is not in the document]
javierdelapuente: I believe there are some issues with the current approach:
 - There are about 60 lines of code copied in a patch, exactly as in the original function, for not putting a 'N\n'. 
 - The new function is also using the standard input in a hacky way (so no much gain, I would rather read stdin than the current way).
 - Besides, the new Ruby function also contains the functionality to update the password in the same way as before (I believe that was one of the reasons for the PR).

Rereading the PR description, it says `Created an add_user action`. However I believe no action was created.

actions.yaml
```diff
@@ -1,19 +1,29 @@
...
+      type: boolean
+      description: Whether the user should be an admin.
+  required: [email]
 anonymize-user:
   description: Anonymize a user.
   params:
     username:
       type: string
       description: The unique identifier of the user to anonymize.
+    admin:
```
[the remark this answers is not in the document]
codethulu: this is erroneous, let me remove this

nrobinaubertin (dismissed):

discourse_rock/patches/discourse-charm.patch
```diff
@@ -0,0 +1,18 @@
...
++
++desc "Check if a user exists for given email address"
++task "users:exists", [:email] => [:environment] do |_, args|
++  email = args[:email]
++  if !email || email !~ /@/
++    puts "ERROR: Expecting rake users:exists[some@email.com]"
++    exit 1
++  end
++  if User.find_by_email(email)
++    puts "ERROR: User with email #{email} not found"
```
javierdelapuente, line 14: This line does the opposite of what is expected, the previous condition is if the user exists and then it states that the user does not exists.

tests/integration/test_saml.py
```diff
@@ -0,0 +1,102 @@
...
+    saml_helper = setup_saml_config
+    urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+    # discourse need a long password and a valid email
+    # username can't be "discourse" or it will be renamed
+    username = "ubuntu"
+    email = "ubuntu@canonical.com"
+    password = "test-discourse-k8s-password"  # nosecue
+    saml_helper.register_user(username=username, email=email, password=password)
+
+    action_result = await run_action(app.name, "create-user", email=email, admin=True)
```
javierdelapuente, line 41: It is interesting that this test is passing. As seen in the file `discourse_rock/patches/discourse-charm.patch`, this means that the rake task `users:exists` is failing someway, otherwise this would not work. 

As a corollary, probably the promote action does not work.

discourse_rock/patches/discourse-charm.patch
```diff
@@ -0,0 +1,18 @@
...
++    exit 1
```
javierdelapuente, line 15: At the moment there is no way of differentiating user not found from a failure (except matching on the exact error string).

I believe this should be handled some other way, let's say using another status code for `user not found` of having status code 0 for `user not found` and returning some meaningful message in stdout (true/false or something like that).

tests/unit/test_charm.py
```diff
@@ -344,13 +344,51 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
+    )
+
+    charm: DiscourseCharm = typing.cast(DiscourseCharm, harness.charm)
+
+    email = "sample@email.com"
+    event = MagicMock(spec=ActionEvent)
+    event.params = {
+        "email": email,
+    }
+    charm._on_create_user_action(event)
```
javierdelapuente, line 391: create-user calls rake twice. I believe this should be checked. In the case of checking if the user exists it should be checked/mocked. What is really happening when this tests runs with the rake task `users:exists`?

tests/unit/test_charm.py
```diff
@@ -344,13 +344,51 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
+    arrange: an email and a password
+    act: when the _on_create_user_action method is executed
+    assert: the underlying rake command to add the user is executed
+        with the appropriate parameters.
+    """
+    harness = helpers.start_harness()
+
+    # We catch the exec call that we expect to register it and make sure that the
+    # args passed to it are correct.
+    expected_exec_call_was_made = False
```
javierdelapuente, line 365: Is this variable used? I cannot see the asset that checks that this variable is true.

tests/unit/test_charm.py
```diff
@@ -344,13 +344,51 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
 
     email = "sample@email.com"
-    password = "somepassword"  # nosec
     event = MagicMock(spec=ActionEvent)
     event.params = {
         "email": email,
-        "password": password,
     }
-    charm._on_add_admin_user_action(event)
+    charm._on_promote_user_action(event)
```
javierdelapuente, line 351: I cannot see the assert that checks that the rake command was called. I believe that something should be also checked/mocked with the rake task `users:exists`?

src/charm.py
```diff
@@ -699,39 +702,112 @@ def _activate_charm(self) -> None:
...
+            [os.path.join(DISCOURSE_PATH, "bin/bundle"), "exec", "rake", "users:exists", email],
+            working_dir=DISCOURSE_PATH,
+            user=CONTAINER_APP_USERNAME,
+            environment=self._create_discourse_environment_settings(),
+        )
+        try:
+            user_exists.wait_output()
+            event.fail(f"User with email {email} already exists")
+        except ExecError:
+            pass
```
javierdelapuente, line 775: I believe we should differentiate between failure and user does not exist.

src/charm.py
```diff
@@ -699,39 +702,112 @@ def _activate_charm(self) -> None:
...
+        user_exists = container.exec(
+            [os.path.join(DISCOURSE_PATH, "bin/bundle"), "exec", "rake", "users:exists", email],
+            working_dir=DISCOURSE_PATH,
+            user=CONTAINER_APP_USERNAME,
+            environment=self._create_discourse_environment_settings(),
+        )
+        try:
+            user_exists.wait_output()
+        except ExecError:
+            event.fail(f"User with email {email} does not exist")
```
javierdelapuente, line 727: I believe we should differentiate between failure and user does not exist.

javierdelapuente (commented):
Reviewing the code, I believe more tests are needed:
- Integration tests that test create-user and promote-user, because currently is the only way of testing the Ruby code.
- Unit tests that test also all the relevant code paths (like user found and promote-user failing because of that).

tests/integration/test_saml.py
codethulu: found why the test is passing, an error was being caused in some other way. a different return code has been implemented, whcih could have caught this

tests/unit/test_charm.py
codethulu: hmm, seems like previous tests did not assert this variable. i have made sure to assert this now.

src/charm.py
merkata: @codethulu can you post a short comment before resolving a conversation? that applies for all the resolved convos. Thanks!

src/charm.py
merkata: :+1:  on what @javierdelapuente is saying, and also - I believe also that you might be getting an ExecError with the email existing and that `pass` is effectively masking the `exit 1` you have, so that might be the reason your integration test is failing...

merkata: :+1:  on what @javierdelapuente is saying, and also - I believe also that you might be getting an ExecError with the email existing and that `pass` is effectively masking the `exit 1` you have, so that might be the reason your integration test is failing...

merkata (commented):
left some comments, and as a general note I would like to have comments in a conversation that has been resolved

tests/integration/test_charm.py
```diff
@@ -466,3 +359,26 @@ def _upgrade_finished():
...
     await model.block_until(upgrade_finished(), timeout=10 * 60)
     check_alive()
+
+
+@pytest.mark.asyncio
+async def test_create_user(
+    app: Application,
+):
+    """
+    arrange: Given discourse application
```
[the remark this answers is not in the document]
codethulu: I have changed the description for this test, and other instances of this in other test descriptions

tests/integration/test_charm.py
```diff
@@ -395,21 +393,57 @@ async def test_promote_user(
     act: when promoting a user
     assert: the user should be promoted successfully.
     """
+    with requests.session() as sess:
+        res = sess.get(f"{discourse_address}/session/csrf", headers={"Accept": "application/json"})
+        # pylint doesn't see the "ok" member
+        assert res.status_code == requests.codes.ok, res.text  # pylint: disable=no-member
```
arturo-seijas, line 399: Why not res.ok?

tests/integration/test_charm.py
```diff
@@ -395,21 +393,57 @@ async def test_promote_user(
     act: when promoting a user
     assert: the user should be promoted successfully.
     """
+    with requests.session() as sess:
+        res = sess.get(f"{discourse_address}/session/csrf", headers={"Accept": "application/json"})
```
arturo-seijas, line 397: s/res/response would increase readability

tests/integration/test_charm.py
```diff
@@ -395,21 +393,57 @@ async def test_promote_user(
...
+        # pylint doesn't see the "ok" member
+        assert res.status_code == requests.codes.ok, res.text  # pylint: disable=no-member
+        data = res.json()
+        assert data["csrf"], data
+        csrf = data["csrf"]
+
+        def get_api_key() -> str:
+            res = sess.post(
```
arturo-seijas, line 405: s/sess/session

tests/integration/test_charm.py
codethulu: similar code can be found in conftest, i decided to keep it consistent with what already exists in the repo

tests/integration/test_charm.py
```diff
@@ -395,21 +393,57 @@ async def test_promote_user(
...
+                    "Content-Type": "application/json",
+                    "X-CSRF-Token": csrf,
+                    "X-Requested-With": "XMLHttpRequest",
+                },
+                json={"key": {"description": "admin-api-key", "username": None}},
+            )
+            return res.json()["key"]["key"]
+
+        def attempt_login(email: str, password: str) -> str:
+            res = sess.post(
```
arturo-seijas, line 417: s/res/response

tests/integration/test_charm.py
```diff
@@ -395,21 +393,57 @@ async def test_promote_user(
...
-            "Api-Key": admin_api_key,
-            "Api-Username": admin_credentials.username,
-        },
-        json=category_info,
-        timeout=60,
-    )
+        attempt_login(email, action.results["password"])
+
+        # This should fail as the user is not promoted
+        assert get_api_key() is None
```
arturo-seijas, line 442: You can use `assert, condition, message` instead and remove the comment explaining the assert

tests/integration/test_charm.py
```diff
@@ -395,21 +393,57 @@ async def test_promote_user(
...
+
+        # Promote the user
+        action: Action = await discourse_unit.run_action("promote-user", email=email)
+        await action.wait()
+
+        # Check the user is promoted
```
arturo-seijas, line 448: Same here

actions.yaml
```diff
@@ -1,15 +1,22 @@
 # Copyright 2024 Canonical Ltd.
 # See LICENSE file for licensing details.
-add-admin-user:
-  description: Add a new admin user.
+promote-user:
```
arturo-seijas, line 3: Can you order these alphabetically?

discourse_rock/patches/discourse-charm.patch
```diff
@@ -0,0 +1,15 @@
+--- a/lib/tasks/discourse-charm.rake
++++ b/lib/tasks/discourse-charm.rake
+@@ -0,0 +1,12 @@
++# frozen_string_literal: true
++
++desc "Check if a user exists for given email address"
```
arturo-seijas, line 6: Can you add a comment here to the upstream PR so that we can keep track of it?

tests/integration/test_saml.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
```
arturo-seijas, line 1: This change is out of the scope of the PR. My suggestion is that, in the future, you open a different PR per change in order to:
* Make reviews easier
* Shorten the feedback cycle, and increasing delivery
* Easier git operations (comparisons/reverts/etc) across revisions

tests/integration/test_charm.py
arturo-seijas: Makes sense, but I'd rather have it all changed incrementally

src/charm.py
```diff
@@ -699,39 +702,113 @@ def _activate_charm(self) -> None:
...
+        )
+        try:
+            process.wait_output()
+            event.set_results({"user": email})
+        except ExecError as ex:
+            event.fail(
+                f"Failed to make user with email {email} an admin: {ex.stdout}"  # type: ignore
             )
+
+    def _on_create_user_action(self, event: ActionEvent, skip_validation: bool = False) -> None:
```
arturo-seijas, line 751: Where is `skip_validation` used?

tests/integration/test_charm.py
codethulu: i have changed everywhere to assert response.ok instead

tests/integration/test_charm.py
codethulu: have done so, and have made consistent across the conftest also.

tests/integration/test_charm.py
```diff
@@ -395,21 +393,57 @@ async def test_promote_user(
...
-        },
-        json=category_info,
-        timeout=60,
-    )
+        attempt_login(email, action.results["password"])
+
+        # This should fail as the user is not promoted
+        assert get_api_key() is None
+
+        # Promote the user
```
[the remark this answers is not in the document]
codethulu: removed

actions.yaml
codethulu: have done so

src/charm.py
codethulu: When we mock in the unit test, all container exec calls fire without failing, even if the rake task fails. 

In this test, we rely on the `users:exists` rake call to fail, to proceed with the user creation. Because this always passes when mocked, the function exits early and the rake command to create the user never occurs. Thus, we need to have some option of skipping validation in order for the unit test to verify the rake command works.

The `create_user` action is tested fully, with more rigour, in the integration test that accompanies it. Unfortunately these rake methods cannot be tested so simply with unit tests

arturo-seijas: Can't you just patch the `.wait_output` return value instead? It'd be better not to have source code that's only used for testing

tests/integration/test_charm.py
codethulu: This has been resolved.

codethulu: (and has been applied elsewhere, where relevant.)

tests/integration/test_charm.py
codethulu: addressed.

tests/integration/test_charm.py
codethulu: addressed.

tests/integration/test_saml.py
codethulu: Yes, this PR has ended up ballooning with a few issues/inconsistencies having been highlighted through the development process. Next time I will try split into smaller, more focussed PRs.

discourse_rock/patches/discourse-charm.patch
codethulu: https://github.com/discourse/discourse/pull/27163

src/charm.py
codethulu: patch added

discourse_rock/patches/discourse-charm.patch
javierdelapuente: I still think the issue is there. If this task fails for whatever reason, the event create user will fail with:
`event.fail(f"User with email {email} does not exist")`

src/charm.py
javierdelapuente: This is still an issue, unless there is a reason I think we should tackle it.

tests/integration/test_users.py
```diff
@@ -0,0 +1,124 @@
...
+    assert: User cannot access the admin API before being promoted
+    """
+
+    def test_discourse_srv_status_ok():
+        response = requests.get(f"{discourse_address}/srv/status", timeout=requests_timeout)
+        assert response.status_code == 200
+
+    # The charm should be active when starting this test
+    await model.wait_for_idle(status="active")
+    test_discourse_srv_status_ok()
```
javierdelapuente, line 62: Not sure why this is needed. 

In the previous test this check is not there. If this test needs it, the previous one also needs it. I would say that none of them needs it. If the unit is active, we should be able to run the test.

docs/how-to/contribute.md
```diff
@@ -83,8 +83,7 @@ the registry:
 
 ```shell
     cd [project_dir]/discourse_rock && rockcraft pack rockcraft.yaml
-    skopeo --insecure-policy copy oci-archive:discourse_1.0_amd64.rock docker-daemon:localhost:32000/discourse:latest
-    docker push localhost:32000/discourse:latest
+    skopeo --insecure-policy copy --dest-tls-verify=false oci-archive:discourse_1.0_amd64.rock docker://localhost:32000/discourse:latest
```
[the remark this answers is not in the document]
codethulu: ah i didn't know, this is a nice change. done!

tests/integration/test_users.py
codethulu: i don't think either needs it, it just seemed as though it had fixed some saml issues yesterday but this was not the case

discourse_rock/patches/discourse-charm.patch
codethulu: i have addressed this now, by returning a different error code and managing this in the except

src/charm.py
codethulu: have addressed this, by returning a different error code

src/charm.py
codethulu: https://github.com/canonical/discourse-k8s-operator/pull/241/files/88bcccafb9209796e2c11248e154cf5338950895#r1638697648

src/charm.py
```diff
@@ -694,39 +697,148 @@ def _activate_charm(self) -> None:
...
+        except ExecError as ex:
+            event.fail(
+                f"Failed to make user with email {email} an admin: {ex.stdout}"  # type: ignore
             )
-            try:
-                process.wait_output()
-                event.set_results({"user": f"{email}"})
-            except ExecError as ex:
+
+    def _on_create_user_action(self, event: ActionEvent) -> None:
```
javierdelapuente, line 751: The logic of this function is simple, but the function is very long and so it is difficult to see the real logic. I believe splitting it into several functions will help with readability. This will imply moving the code that calls pebble to its own functions.

src/charm.py
```diff
@@ -694,39 +697,148 @@ def _activate_charm(self) -> None:
             self._start_service()
             self.model.unit.status = ActiveStatus()
 
-    def _on_add_admin_user_action(self, event: ActionEvent) -> None:
-        """Add a new admin user to Discourse.
+    def _on_promote_user_action(self, event: ActionEvent) -> None:
+        """Promote a user to a specific trust level.
```
javierdelapuente, line 701: The logic of this function is simple, but the function is very long and so it is difficult to see the real logic. I believe splitting it into several functions will help with readability. This will imply moving the code that calls pebble to its own functions.

tests/integration/test_users.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+@pytest.mark.asyncio
+async def test_create_user(
+    app: Application,
+):
+    """
+    arrange: A discourse application
+    act: Create a user
+    assert: User is created, and re-creating the same user should fail
```
javierdelapuente, line 24: This test is testing that when the action is run twice with a username, it fails. It is not really testing that a user is created. I believe a real assertion that the user is created could help.

tests/unit/test_charm.py
```diff
@@ -344,13 +345,86 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
+
+    mock_exec.side_effect = mock_exec_side_effect
+    mock_exec.return_value = MagicMock(
+        wait_output=mock_wo,
+    )
+
+    charm: DiscourseCharm = typing.cast(DiscourseCharm, harness.charm)
+    event = MagicMock(spec=ActionEvent)
+    event.params = {"email": email}
+    charm._on_create_user_action(event)
```
javierdelapuente, line 426: You can run `harness.run_action` instead of calling a private method, making the test more realistic.

tests/unit/test_charm.py
```diff
@@ -344,13 +345,86 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
+def test_create_user(mock_exec):
+    """
+    arrange: an email
+    act: when the _on_create_user_action method is executed
+    assert: the create user rake command is executed upon failure of the user existence check.
+    """
+    mock_wo = MagicMock(
+        return_value=("", None),
+    )
+    stdout_mock = "CRASH"
```
javierdelapuente, line 364: This should be adapted to the real stdout of user not found.

tests/unit/test_charm.py
```diff
@@ -344,13 +345,86 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
-        "email": email,
-        "password": password,
-    }
-    charm._on_add_admin_user_action(event)
+    event.params = {"email": email}
+    charm._on_promote_user_action(event)
+    assert expected_exec_call_was_made
+
+
+@patch.object(ops.Container, "exec")
```
javierdelapuente, line 354: the ops library already has `harness.handle_exec`. This function has clear interfaces and is easy to patch in a similar way to a real exec of pebble. There is no need to patch the exec in the container class.

tests/unit/test_charm.py
```diff
@@ -344,13 +345,86 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
+        return_value=("", None),
+    )
+    stdout_mock = "CRASH"
+    harness = helpers.start_harness(run_initial_hooks=False)
+    harness.set_can_connect(CONTAINER_NAME, True)
+
+    email = "admin-user@test.internal"
+    expected_exec_call_was_made = False
+
+    def mock_wo_side_effect(*args, **kwargs):  # pylint: disable=unused-argument
```
javierdelapuente, line 371: This mocking could be substituted with something like:
```
    harness.handle_exec(
        SERVICE_NAME,
        [f"{DISCOURSE_PATH}/bin/bundle", "exec", "rake", f"users:exists[{email}]"],
        result=ops.testing.ExecResult(exit_code=2, stdout="not found", stderr="")
    )
    ```

File: tests/unit/test_charm.py [hunk 44]
```diff
@@ -344,13 +345,86 @@ def bundle_handler(args: ops.testing.ExecArgs) -> None:
...
+        """
+
+        if isinstance(mock_wo.cmd, list) and f"users:exists[{email}]" in mock_wo.cmd:
+            raise ExecError(command=mock_wo.cmd, exit_code=2, stdout=stdout_mock, stderr="")
+
+        return DEFAULT
+
+    mock_wo.side_effect = mock_wo_side_effect
+
+    def mock_exec_side_effect(*args, **kwargs):  # pylint: disable=unused-argument
```
javierdelapuente, line 392: This can be updated to be something like:
```
    mock_create_user = unittest.mock.Mock(
        return_value=ops.testing.ExecResult(exit_code=0, stdout="ok, created", stderr="")
    )
    harness.handle_exec(
        SERVICE_NAME,
        [f"{DISCOURSE_PATH}/bin/bundle", "exec", "rake", "admin:create"],
        handler=mock_create_user,
    )
    ...
    mock_create_user.assert_called_once()
   
```

javierdelapuente (commented):
A few more comments! It is on good track!

discourse_rock/rockcraft.yaml
```diff
@@ -186,7 +186,8 @@ parts:
     after: [discourse, patches]
     override-stage: |
       git -C srv/discourse/app apply patches/lp1903695.patch
-      git -C srv/discourse/app apply patches/anonymize_user.patch
+      touch srv/discourse/app/lib/tasks/discourse-charm.rake
```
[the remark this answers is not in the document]
codethulu: this has been removed, seems to be working fine

src/charm.py
codethulu: I have split the users exist call into a separate method, since this is reusable. I have tried to tidy up the methods, but I don't think moving the code that calls pebble will help much -

it will just create a number of non-reusable methods that greater obfuscate the action behaviour. Also, this would not be consistent with the anonymize user action. 

With the user exists code being moved, the promote action is now roughly the same size as that anonymize action and just as readable in my opinion, and the create action is really only more complex due to the user activation at the end

tests/integration/test_users.py
codethulu: The second call of the action is verifying the user exists, through the rake call, which will only be successful if the user exists. I could include an API call to verify this, but i feel that this verification is covered in the rake task?

tests/unit/test_charm.py
codethulu: Nice to know, I was looking at [indico](https://github.com/canonical/indico-operator/blob/main/tests/unit/test_core.py#L71) for reference. Your suggestion has been implemented!

tests/unit/test_charm.py
codethulu: This has been implemented, much nicer. Thanks

tests/unit/test_charm.py
codethulu: applied your suggestion

tests/unit/test_charm.py
codethulu: I would like to keep consistent with how we verify anonymize user, where we check arguments

tests/unit/test_charm.py
codethulu: This has been implemented now. Thanks :)

tests/unit/test_charm.py
javierdelapuente: An alternative is to check the `mock_calls` in the mock, but that is good too!

tests/integration/test_users.py
javierdelapuente: The assertion in this case consists in running the function that is tested (again) and the test should name should be something like: "test_running_action_create_user_twice_fails". 

But this test is not testing the state of the system, I mean, that there is a user created in discourse.

src/charm.py
javierdelapuente: Separating the `_user_exists` is an improvement! Feel free to close this comment.

My opinion, in general, is that it is not only about being able to reuse functions, but about readability (and when it gets complex, separating by abstractions levels). For example, for create_user, a function like this could separate it from the pebble calls (and the pebble calls are hidden in _user_exists, _create_user and _activate_user):
```
        container = self.unit.get_container(CONTAINER_NAME)
        if not container.can_connect():
            event.fail("Unable to connect to container, container is not ready")
            return

        email = event.params["email"]
        password = self._generate_password(16)
        admin = bool(event.params.get("admin", False))
        active = bool(event.params.get("active", False))

        if self._user_exists(email):
            event.fail(f"User with email {email} already exists")
            return

        self._create_user(email, password, admin)

        if not admin or active:
            self._activate_user(email)

        event.set_results({"user": email, "password": password})
```

src/charm.py
codethulu: this has been addressed in same way as create user

src/charm.py
```diff
@@ -693,40 +696,151 @@ def _activate_charm(self) -> None:
...
+            timeout=60,
+        )
+        try:
+            process.wait_output()
+            event.set_results({"user": email, "password": password})
+        except ExecError as ex:
+            event.fail(f"Failed to make user with email {email}: {ex.stdout}")  # type: ignore
+            return
+
+        if event.params.get("admin") or not event.params.get("active"):
```
nrobinaubertin, line 803: I'm not sure that I get that condition. Why do we skip activation if admin is True ?
Can you add a comment for that ?

nrobinaubertin (commented):
I feel like the PR has out-of-scope changes (like Arturo mentioned). Is it too late to remove some of the changes (in the tests) to move them to another PR ?

src/charm.py
```diff
@@ -693,40 +696,151 @@ def _activate_charm(self) -> None:
...
+            user=CONTAINER_APP_USERNAME,
+            environment=self._create_discourse_environment_settings(),
+        )
+        try:
+            user_exists.wait_output()
+            return True
+        except ExecError as ex:
+            if ex.exit_code == 2:
+                return False
+            raise ex
```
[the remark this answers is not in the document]
codethulu: done!

javierdelapuente (approved):

nrobinaubertin (approved):

tests/integration/test_users.py
codethulu: the test checks a user is created via the rake task

testing user creation is also verified in the following method

PR #278, opened 2024-08-14
feat: Discourse 3.3.0 upgrade

alithethird:
Updating to Discourse v3.3.0

Applicable spec: N/A

### Overview
Upgrading the Discourse version from v3.2.0 to v.3.3.0
Added new part to install `imagemagick` from source because the used `imagemagick` is bumped to v7.1.1. Previously it was installed by apt and the version there is 6.9.

### Checklist


discourse_rock/patches/sigterm.patch
```diff
@@ -15,14 +15,14 @@
...
- 
+
  after_fork do |server, worker|
 +  Signal.trap 'TERM' do
 +    puts 'Unicorn worker intercepting TERM and doing nothing. Wait for master to sent QUIT'
 +  end
    DiscourseEvent.trigger(:web_fork_started)
    Discourse.after_fork
- end
+   SignalTrapLogger.instance.after_fork
```
merkata, line 28: I'm seeing you put this here as v3.3.0 added a SignalTrapLogger and it is used because now they are trapping `USR2`. Perhaps it's better to place the patch nearer the other trap, maybe right bellow it?

v3.2.0
```ruby
after_fork do |server, worker|
  DiscourseEvent.trigger(:web_fork_started)
  Discourse.after_fork
end
```

v3.3.0
```ruby
after_fork do |server, worker|
  DiscourseEvent.trigger(:web_fork_started)
  Discourse.after_fork
  SignalTrapLogger.instance.after_fork

  Signal.trap("USR2") do
    message = <<~MSG
    Unicorn worker received USR2 signal indicating it is about to timeout, dumping backtrace for main thread
    #{Thread.current.backtrace&.join("\n")}
    MSG

    SignalTrapLogger.instance.log(Rails.logger, message, level: :warn)
  end
end
```

wdyt?

discourse_rock/rockcraft.yaml
```diff
@@ -280,6 +279,108 @@ parts:
       chown -R 584792:584792 var/lib/pebble/default/.npm
       chown -R 584792:584792 var/lib/pebble/default/.yarn
       chown    584792:584792 var/lib/pebble/default/.yarnrc
+  imagemagick:
+    plugin: autotools
+    source: https://github.com/ImageMagick/ImageMagick.git
+    source-type: git
+    source-tag: 7.1.1-36
+    build-packages:
```
merkata, line 287: This is confirmed to work, though I don't think we need the same mix of deps when building and when staging. I would expect a build to require the library headers of `libde265` (so `libde265-dev`) and the staged artefact to link to `libde265-0`. It's a minor thing, though you could look into it. A quick refactor would be to put the `-dev` packages in the build (and remove them from stage) and make sure the rock gets built just fine and Discourse is working (I expect it would). You don't need to do this as part of this PR..

merkata (dismissed):
This looks good, great work @alithethird ! I left two comments, they are about tidying up, so you can address them at a later point.

discourse_rock/rockcraft.yaml
```diff
@@ -280,6 +279,108 @@ parts:
       chown -R 584792:584792 var/lib/pebble/default/.npm
       chown -R 584792:584792 var/lib/pebble/default/.yarn
       chown    584792:584792 var/lib/pebble/default/.yarnrc
+  imagemagick:
```
amandahla, line 282: Could you add a note in the PR description about why are you adding this imagemagick now please?

alithethird: Good point, thanks!

discourse_rock/rockcraft.yaml
alithethird: Working on it.

discourse_rock/patches/sigterm.patch
alithethird: Are you proposing this change to keep the code prettier? By putting traps back to back. Or does it serve a functional purpose?

So are you saying its better to do it like this?

```
after_fork do |server, worker|
  DiscourseEvent.trigger(:web_fork_started)
  Discourse.after_fork
  SignalTrapLogger.instance.after_fork

  Signal.trap("USR2") do
    message = <<~MSG
    Unicorn worker received USR2 signal indicating it is about to timeout, dumping backtrace for main thread
    #{Thread.current.backtrace&.join("\n")}
    MSG

    SignalTrapLogger.instance.log(Rails.logger, message, level: :warn)
  end
  Signal.trap 'TERM' do
    puts 'Unicorn worker intercepting TERM and doing nothing. Wait for master to sent QUIT'
  end
end
```

discourse_rock/rockcraft.yaml
alithethird: I removed the `-dev` packages from `stage-packages` and it doesn't work unfortunately. I made sure they are still in `build-packages`.
I will work more on it tomorrow.

nrobinaubertin (dismissed):

nrobinaubertin (approved):

discourse_rock/patches/sigterm.patch
merkata: It's purely cosmetical and up to you.

merkata (approved):
lgtm

Issue #268, opened 2024-07-02
Redis relation does not read all sentinel addresses


reneradoi:
### Bug Description

Hi Team!

In `redis-k8s-operator` we run an integration test deploying and relating `discourse-k8s` to Redis. On some CI runs I experienced issues with the discourse charm because it can't write to the Redis database.

The reason for this is that the `redis-client` tries to connect to the first advertised host it can find in the relation data ([here](https://github.com/canonical/discourse-k8s-operator/blob/main/src/charm.py#L545-L549)). But since a bug was fixed on `redis-k8s-operator` side, the relation data now includes all hosts there, not just the one of the Sentinel Master anymore. This was fixed [with this PR](https://github.com/canonical/redis-k8s-operator/pull/89), which means you now have to consider all advertised hosts, not just anyone anymore.

One more notice: This behaviour only happens if the Sentinel Master is not the unit `/0`. If, by accident, `redis-k8s/0` is the Sentinel Master, or in setups with just one Redis unit, the integration works fine.

### To Reproduce

1. `juju deploy redis-k8s --channel edge -n 3`
2. `juju deploy postgresql-k8s --channel 14/stable`
3. `juju deploy discourse-k8s --channel latest/stable`
4. `juju integrate postgresql-k8s:database discourse-k8s`
5. `juju integrate redis-k8s discourse-k8s`

### Environment

see above.

Example for failed CI run: https://github.com/canonical/redis-k8s-operator/actions/runs/9760370147/job/26939124902

### Relevant log output

```shell
from `juju debug-log`:


unit-discourse-k8s-0: 11:35:26 ERROR unit.discourse-k8s/0.juju-log redis:11: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 805, in <module>
    main(DiscourseCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-discourse-k8s-0/charm/venv/ops/main.py", line 544, in main
    manager.run()
  File "/var/lib/juju/agents/unit-discourse-k8s-0/charm/venv/ops/main.py", line 520, in run
    self._emit()
  File "/var/lib/juju/agents/unit-discourse-k8s-0/charm/venv/ops/main.py", line 509, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-discourse-k8s-0/charm/venv/ops/main.py", line 143, in _emit_charm_event
[...]

PR #283, opened 2024-09-03
Feat/ Redis HA fix

alithethird:
Includes the [Redis HA fix](https://github.com/canonical/redis-k8s-operator/pull/96). This fixes 2 different bugs: 
1- https://bugs.launchpad.net/juju/+bug/2063087
2- https://github.com/canonical/discourse-k8s-operator/issues/268

### Overview

Started using app databag for Redis hostname. Add tests to test out Redis in HA mode. 

### Rationale

Previously was not able to use Redis in HA mode

### Checklist


src/charm.py
```diff
@@ -390,16 +390,17 @@ def _get_redis_relation_data(self) -> typing.Tuple[str, int]:
...
-        relation_data = self.redis.relation_data
-        if not relation_data:
+        relation = self.model.get_relation(self.redis.relation_name)
+        if not relation:
             raise MissingRedisRelationDataError("No redis relation data")
 
         try:
-            redis_hostname = str(relation_data["hostname"])
-            redis_port = int(relation_data["port"])
+            relation_data = relation.data[self.model.relations["redis"][0].app]
```
javierdelapuente, line 398: Maybe something like this:
`relation_data_app = relation.data[relation.app]`.

javierdelapuente (commented):
One question, what are now the restrictions to integrate with Redis? A specific version?

Should we fallback to the previous functionality if "leader-host" is not there? Otherwise, should we tell the user that this charm requires a Redis charm revision greater than X?

alithethird:
> One question, what are now the restrictions to integrate with Redis? A specific version?
> 
> Should we fallback to the previous functionality if "leader-host" is not there? Otherwise, should we tell the user that this charm requires a Redis charm revision greater than X?

Ok, that is a great question. I did not think of that. I think I should ask @gregory-schiano about this. 

I think the fallback to unit data if `leader-host` doesn't exist is a great idea. What do you think?

src/charm.py
```diff
@@ -390,16 +390,23 @@ def _get_redis_relation_data(self) -> typing.Tuple[str, int]:
...
-            redis_hostname = str(relation_data["hostname"])
-            redis_port = int(relation_data["port"])
+            if not redis_hostname:
+                redis_hostname = str(self.redis.relation_data["hostname"])
+            redis_port = int(self.redis.relation_data["port"])
         except (KeyError, ValueError) as exc:
             raise MissingRedisRelationDataError(
-                "Wrong hostname or port in Redis relation"
+                "Wrong hostname or port in Redis App data"
             ) from exc
```
gregory-schiano, line 410: The whole structure of the code bothers me a bit, can't we do something like:
```python
try:
   redis_port = int(relation_data["port"])
   redis_hostname = str(relation_app_data["leader-host"] if relation_app_data.get("leader-host") else relation_app_data["hostname"])
except (KeyError, ValueError) as exc:
   raise MissingRedisRelationDataError("Either 'leader-host' or 'hostname' and 'ports' are mandatory") from exc
```

Or something around the lines

alithethird: Sure can do! I will also do the same in Indico Charm.

src/charm.py
```diff
@@ -390,16 +390,22 @@ def _get_redis_relation_data(self) -> typing.Tuple[str, int]:
...
+        relation = self.model.get_relation(self.redis.relation_name)
+        if not relation:
             raise MissingRedisRelationDataError("No redis relation data")
+        relation_app_data = relation.data[relation.app]
+        relation_unit_data = self.redis.relation_data
 
         try:
-            redis_hostname = str(relation_data["hostname"])
-            redis_port = int(relation_data["port"])
+            redis_hostname = str(
```
javierdelapuente, line 400: As an alternative:

`redis_hostname = str(relation_app_data.get("leader-host", relation_unit_data["hostname"]))`

All the catching of the exceptions is a bit strange, but can be refactored in the future as it is not what this change did.

javierdelapuente (approved):

src/charm.py
alithethird: Thats even cleaner! Thank you :D. I will remember this if I do a refactor pr in the future.

amandahla (approved):
LGTM

PR #287, opened 2024-09-10
Feat: Fixed db migration issues from v3.2.0 to 3.3.0

alithethird:
When migrating from v3.2.0 to v3.3.0 we had to drop some triggers by hand. This PR makes it so that the triggers are dropped automagically.

```
DROP TRIGGER IF EXISTS invites_user_id_readonly ON invites;

DROP TRIGGER IF EXISTS invites_redeemed_at_readonly ON invites;

DROP TRIGGER IF EXISTS user_api_keys_scopes_readonly ON user_api_keys;
```
### Checklist


discourse_rock/patches/db_migrations.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/db/post_migrate/20240212034010_drop_deprecated_columns.rb b/db/post_migrate/20240212034010_drop_deprecated_columns.rb
```
arturo-seijas, line 1: Have you contributed this change upstream too?

arturo-seijas (dismissed):

discourse_rock/patches/db_migrations.patch
arturo-seijas: Yes, please :) https://github.com/discourse/discourse

alithethird: Sure :D but I think my fix is pretty janky :sweat_smile:

arturo-seijas: :D Nah, I think it's fine :)

charmcraft.yaml
```diff
@@ -12,5 +12,6 @@ bases:
       channel: "20.04"
 parts:
   charm:
+    build-packages: [cargo, rustc, pkg-config, libffi-dev, libssl-dev]
```
arturo-seijas, line 15: I think you can remove `cargo`, `rust` and `pkg-config` by including
```
build-snaps:
      - rustup
```
instead

alithethird: oh cool, let me try that. Thanks!

alithethird: That did not work actually.

`error: can't find Rust compiler`
log:
https://pastebin.canonical.com/p/hjj7kVzzs8/

discourse_rock/patches/db_migrations.patch
```diff
@@ -0,0 +1,17 @@
...
+index 0899da20..586643ba 100644
+--- a/db/post_migrate/20240212034010_drop_deprecated_columns.rb
++++ b/db/post_migrate/20240212034010_drop_deprecated_columns.rb
+@@ -19,6 +19,13 @@ class DropDeprecatedColumns < ActiveRecord::Migration[7.0]
+   }
+
+   def up
++    execute <<~SQL
++    DROP TRIGGER IF EXISTS invites_user_id_readonly ON invites;
```
amandahla, line 10: Out of curiosity: in the case of downgrade/roll back, are the triggers recreated?

charmcraft.yaml
arturo-seijas: rustc and cargo should be included, as per https://www.rust-lang.org/tools/install. Can you try keeping pkg-config and cleaning the charmcraft cache just in case?

discourse_rock/patches/db_migrations.patch
alithethird: No, because also the columns are not recreated. 
The following is the downgrading function:
```ruby
  def down
    raise ActiveRecord::IrreversibleMigration
  end
```

charmcraft.yaml
alithethird: Sure, trying it out

alithethird: Tried:
```python

  charm:
    build-packages: [cargo, rustc, pkg-config, libffi-dev, libssl-dev]
    build-snaps:
      - rustup
```
and 
```python

  charm:
    build-packages: [cargo, rustc, pkg-config]
    build-snaps:
      - rustup
```

Both did not work.

arturo-seijas: Well, let's keep it as it is then

discourse_rock/patches/db_migrations.patch
amandahla: In that case the upgrade document should be updated with a warning about creating a database backup.
Similar to what Wordpress has
https://github.com/canonical/wordpress-k8s-operator/blob/main/docs/how-to/upgrade-wordpress-charm.md

tests/integration/test_charm.py
```diff
@@ -78,6 +78,88 @@ async def test_setup_discourse(
...
+    return_code, _, _ = await ops_test.juju(
+        "ssh",
+        "--container",
+        "postgresql",
+        postgres_app.units[0].name,
+        "pg_restore -h localhost -U operator\
+              --password -d discourse\
+                  --no-owner --clean --if-exists ./mock_db",
+        stdin=str.encode(f"{db_pass}\n"),
+    )
```
mthaddon, line 138: We're loading `./mock_db` which is a binary file, which makes it very hard to inspect/verify the contents. Would it either be possible to have that file be in a non-binary format, or can we do some verification in this test that the database is in the state we want it to be in (e.g. that it's from Discourse v3.2.0)?

tests/integration/test_charm.py
```diff
@@ -78,6 +78,88 @@ async def test_setup_discourse(
     assert response.status_code == 200
 
 
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_db_migration(model: Model, ops_test: OpsTest, pytestconfig: Config, run_action):
+    """
+    arrange: preload postgres with a mock db
+    act: deploy and integrate with discourse
+    assert: discourse must be active idle
```
mthaddon, line 87: I feel like this isn't quite describing what we're doing fully. We load a v3.2.0 DB which has contents that would trigger a bug (can we mention what bug as a reference here), and then we deploy discourse v3.3.0 and verify that it handles the schema migration correctly.

tests/integration/test_charm.py
```diff
@@ -78,6 +78,88 @@ async def test_setup_discourse(
...
+            "plugin_pg_trgm_enable": "true",
+        }
+    )
+    await model.wait_for_idle(apps=[postgres_app.name], status="active")
+    db_pass = await run_action(postgres_app.name, "get-password", username="operator")
+    db_pass = db_pass["password"]
+    return_code, _, _ = await ops_test.juju(
+        "scp",
+        "--container",
+        # postgres_app.name,
```
mthaddon, line 112: Is this line needed?

tests/integration/test_charm.py
alithethird: Thanks for pointing that out. It looks like I can do `pg_restore db.bin > db.sql` to make it human readable but I need to test if it is still importable in `.sql` format. For verification I can first install Discourse v3.2.0 to show that it works with the db then upgrade it to the current version. I did not do this to keep the test shorter and less resource hungry

tests/integration/test_charm.py
alithethird: I can write it in a more detailed way, thank you!

tests/integration/test_charm.py
alithethird: Not needed, I missed it when cleaning the code. Thanks for pointing it out!

tests/integration/test_db_migration.py
```diff
@@ -0,0 +1,95 @@
...
+    arrange: preload postgres with a mock db that is created in Discource v3.2.0
+    act: deploy and integrate with discourse v3.3.0
+    assert: discourse must be active idle, previously it was creating migration
+    errors related to not being able to delete some columns because of triggers
+    """
+    postgres_app = await model.deploy(
+        "postgresql-k8s",
+        channel="14/stable",
+        series="jammy",
+        revision=300,
```
mthaddon, line 29: Do we really need to specify a particular revision here?

tests/integration/test_db_migration.py
```diff
@@ -0,0 +1,95 @@
...
+from pytest_operator.plugin import Model, OpsTest
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_db_migration(model: Model, ops_test: OpsTest, pytestconfig: Config, run_action):
+    """
+    arrange: preload postgres with a mock db that is created in Discource v3.2.0
```
mthaddon, line 20: s/Discource/Discourse/

tests/integration/test_db_migration.py
```diff
@@ -0,0 +1,95 @@
...
+    act: deploy and integrate with discourse v3.3.0
```
mthaddon, line 21: Can we capitalise Discourse like we do in the line above?

tests/integration/test_db_migration.py
```diff
@@ -0,0 +1,95 @@
...
+    assert: discourse must be active idle, previously it was creating migration
```
mthaddon, line 22: I'd suggest we just include `assert: discourse is active/idle`, and then after an empty line we have a bit more explanation about how the migrate fails without the drop triggers patch.

testing_database/creating-the-testing-database.md
```diff
@@ -0,0 +1,34 @@
+# Creating the testing database
+At the writing this document, the testing database is created using Discourse v3.2.0.
+
+## Creating the database
+First of all we need to deploy the Discourse following the [tutorial](https://github.com/canonical/discourse-k8s-operator/blob/main/docs/tutorial.md).
```
mthaddon, line 5: When this PR lands, deploying discourse by following the regular tutorial would give us v3.3.0 wouldn't it?

testing_database/creating-the-testing-database.md
```diff
@@ -0,0 +1,34 @@
+# Creating the testing database
```
mthaddon, line 1: I'm not sure if creating a new "testing_database" folder is the right way to go here. Perhaps we could include this as an extended doc string in the test_db_migration integration test? I'd be interested in others feedback on this before we decide which way to go.

tests/integration/test_db_migration.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
```
mthaddon, line 1: Do we need both mock_db and testing_database.sql? I think we only need one, right?

tests/integration/test_db_migration.py
```diff
@@ -0,0 +1,95 @@
...
+    return_code, _, ssh_err = await ops_test.juju(
+        "ssh",
+        "--container",
+        "postgresql",
+        postgres_app.units[0].name,
+        "pg_restore -h localhost -U operator\
+              --password -d discourse\
+                  --no-owner --clean --if-exists ./testing_database.sql",
+        stdin=str.encode(f"{db_pass}\n"),
+    )
```
mthaddon, line 73: I still feel like we're missing something that gives us confidence the database is in the state we want it to be. Is there a "patch" table or "migration" table or something that records what version of the DB we are currently running, and could we query that to confirm that we have the same version as expected for 3.2.0?

tests/integration/test_db_migration.py
alithethird: Oops

tests/integration/test_db_migration.py
alithethird: Ah, forgot to delete the mock_db, good catch!

tests/integration/test_db_migration.py
alithethird: Not really but it was specified in https://github.com/canonical/discourse-k8s-operator/blob/50e229a3ef8e5ac0f8dbed2398a276c7833d2e6b/tests/integration/conftest.py#L155

testing_database/creating-the-testing-database.md
alithethird: Discussed this with @gregory-schiano  on architecture hours and decided to create a new folder but we can discuss again.

tests/integration/test_db_migration.py
alithethird: Searched through the db but was not able to find anything about the version. Maybe we can deploy v3.2.0 with this db to show it doesn't create any errors. Wdyt?

testing_database/creating-the-testing-database.md
alithethird: Changed it to 
`At the writing this document, the testing database is created using Discourse v3.2.0 and used to test migration from v3.2.0 to v3.3.0.`

to make it more clear.

mthaddon: Yes, but as soon as this PR lands this documentation will be obsolete, so I don't think it's a good idea to land it as is.

alithethird: Makes sense, the goal of this documentation is to explain how to create a testing db and the steps are not effected by the version. What should I do in this case?

tests/integration/test_db_migration.py
mthaddon: What about the `schema_migrations` and `schema_migration_details` tables?

tests/integration/test_db_migration.py
mthaddon: Ok, I agree consistency is better. Will close out this comment thread.

tests/integration/test_db_migration.py
alithethird: They specify the migrations that happened that are listed in here: https://github.com/discourse/discourse/tree/main/db/migrate

They don't specify the discourse version but I assume we can get the latest migration file from there and maybe confirm its the v3.2.0

mthaddon (approved):
LGTM, thx!

merkata (approved):
to the moon :rocket:

PR #296, opened 2024-10-03
chore(docs): Update tutorial

alithethird:
### Overview

Standarise the Discourse tutorial with other tutorials and improve/fix the tutorial.

### Checklist

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -10,17 +10,25 @@ In this tutorial, we'll go through each step of the process to get a basic Disco
 
 ## Requirements
 
-* A laptop or desktop running Ubuntu (or you can use a VM).
-* Juju and [Microk8s](https://juju.is/docs/olm/microk8s) installed. We’ll also want to make sure the ingress add-on is enabled, which we can do by running `microk8s enable ingress`.
+- A laptop or desktop running Ubuntu (or you can use a VM).
+- Juju 3 installed and bootstrapped to a MicroK8s controller. You can accomplish this process by following this guide: [Set up / Tear down your test environment](https://juju.is/docs/juju/set-up--tear-down-your-test-environment)
+- NGINX Ingress Controller. If you're using [MicroK8s](https://microk8s.io/), this can be done by running the command `microk8s enable ingress`. For more details, see [Addon: Ingress](https://microk8s.io/docs/addon-ingress).
 
 ## Steps
 
-### Set up environment
+### Shell into the Multipass VM
 
-To easily clean up the resources and separate your workload from the contents of this tutorial, set up a new Juju model named `discourse`:
-
+To be able to work inside the Multipass VM first you need to log in with the following command:
+```bash
+multipass shell my-juju-vm
 ```
-juju add-model discourse
+
+### Add a Juju model for the tutorial
+
+To easily clean up the resources and separate your workload from the contents of this tutorial, set up a new Juju model named `discourse-tutorial`:
+
+```bash
+juju add-model discourse-tutorial
 ```
 
 ### Deploy the charms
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -30,68 +38,67 @@ Discourse requires connections to PostgreSQL and Redis. For more information, se
...
-### Provide ingress capabilities 
+### Provide ingress capabilities
 
 In order to expose the charm, the Nginx Ingress Integrator needs to be deployed and integrated with Discourse:
 
-```
+```bash
 juju deploy nginx-ingress-integrator
-# If your cluster has RBAC enabled you'll be prompted to run the following:
+#If your cluster has RBAC enabled, you'll be prompted to run the following (If you are working inside the Multipass vm, chances are you have RBAC enabled):
```
nrobinaubertin, line 101: I think that adding a way to check for RBAC wouldn't hurt here (`microk8s status`)

docs/tutorial.md
```diff
@@ -100,22 +107,44 @@ juju integrate discourse-k8s nginx-ingress-integrator
...
+ip a | awk '/inet .* ens3/{print $2}' | cut -d'/' -f1
+```
+The result should be something like this:
+```bash
+10.131.49.76
+```
 
+Add the ip address to the /etc/hosts file:
+```bash
+echo "10.131.49.76 discourse-k8s" >> /etc/hosts
```
javierdelapuente, line 127: I would change this to: `echo "10.131.49.76 discourse-k8s" | sudo tee -a /etc/hosts`

docs/tutorial.md
```diff
@@ -100,22 +107,44 @@ juju integrate discourse-k8s nginx-ingress-integrator
...
+```bash
 juju run discourse-k8s/0 create-user admin=true email=email@example.com
 ```
-The command will return the password of the created user. Discourse will be deployed with `discourse-k8s` as default hostname. In order to reach it, modify your `/etc/hosts` file so that it points to `127.0.0.1`:
+The command will return the password of the created user. Discourse will be deployed with `discourse-k8s` as default hostname.
+If you are using a Multipass instance you need to forward the request from your local to the Multipass instance.
+First get the Multipass instances IP address. Since the indico is served on the local address of the Multipass VM we need to use the ip address of the VM. To get the IP address of a Multipass instance run the following command:
+
+```bash
+ip a | awk '/inet .* ens3/{print $2}' | cut -d'/' -f1
```
javierdelapuente, line 118: This will fail in many cases (for example if not using the qemu backend and instead using the lxd backend in multipass). I would use instead:
`ip -4 -j route get 2.2.2.2 | jq -r '.[] | .prefsrc'`

docs/tutorial.md
```diff
@@ -10,17 +10,25 @@ In this tutorial, we'll go through each step of the process to get a basic Disco
 
 ## Requirements
 
-* A laptop or desktop running Ubuntu (or you can use a VM).
-* Juju and [Microk8s](https://juju.is/docs/olm/microk8s) installed. We’ll also want to make sure the ingress add-on is enabled, which we can do by running `microk8s enable ingress`.
+- A laptop or desktop running Ubuntu (or you can use a VM).
```
erinecon, line 13: I would suggestion changing this line to "A working station, e.g., a laptop, with amd64 architecture."

docs/tutorial.md
alithethird: Thank you!!!!

docs/tutorial.md
```diff
@@ -30,92 +39,128 @@ Discourse requires connections to PostgreSQL and Redis. For more information, se
...
 
+```bash
+juju destroy-model discourse-tutorial --destroy-storage
 ```
-juju destroy-model discourse --destroy-storage
+To remove the Multipass instance you created for this tutorial, use the following command.
+```bash
+multipass delete --purge my-juju-vm
 ```
+It is also a good idea to remove the `10.131.49.76 discourse-k8s` line from the `/etc/hosts` file.
```
erinecon, line 166: Nitpick: `It is also a good idea to` --> `Finally,`

erinecon (dismissed):
LGTM

docs/tutorial.md
```diff
@@ -9,18 +9,27 @@
...
+For more information about how to install Juju, see [Get started with Juju](https://juju.is/docs/olm/get-started-with-juju).
 
 ## Steps
 
-### Set up environment
-
-To easily clean up the resources and separate your workload from the contents of this tutorial, set up a new Juju model named `discourse`:
+### Shell into the Multipass VM
 
+To be able to work inside the Multipass VM first you need to log in with the following command:
```
amandahla, line 22: This step is confusing since Multipass is not a requirement.

amandahla: Same for other mentions to Multipass VM.

alithethird: We are adding Multipass VM to the tutorials to simplify it for beginners. @erinecon should we add Multipass as a requirement? The tutorial doesn't necessarily need Multipass but we use Multipass in the tutorial to keep it simple to set up.

erinecon: Multipass is set up if the reader follows the instructions in [Set up / Tear down your test environment](https://juju.is/docs/juju/set-up--tear-down-your-test-environment). So I guess that the requirements can be clearer, i.e., instead of "Juju 3 installed and bootstrapped to a MicroK8s controller" the line could read "A Multipass VM with Juju 3 installed and bootstrapped to a MicroK8s controller."

erinecon: Or, if we want to leave the possibility open to **not** use a Multipass VM, then we could update the requirement to say "Juju 3 installed and bootstrapped to a MicroK8s controller. You can accomplish this process by using a Multipass VM as outlined in this guide: [Set up / Tear down your test environment](https://juju.is/docs/juju/set-up--tear-down-your-test-environment)"

amandahla: I agree with the last suggestion so the warnings about Multipass will make sense when reading the tutorial.

amandahla (approved):
LGTM

erinecon (approved):

PR #317, opened 2025-02-05
Use latest version supporting 3.3

seb4stien:
Upgrading to discourse 3.3 breaks the compatibility with the pinned discourse-markdown-note module.
A new version of the module has been released, and we're updating the dependency accordingly.


gregory-schiano (approved):
Approving to expedite, but can you add a description ?

jdkandersson (approved):

PR #338, opened 2025-06-06
ISD-1817 Add ability to automatically configure cors_origin

### LINKED ISSUE #218: Automatically include `external_hostname` & `s3_cdn_url` in CORS origin (`cors_origin`)
### Enhancement Proposal

Hi,

I think it would be nice to automatically include the configured `external_hostname` & `s3_cdn_url` in the CORS origin (`cors_origin`).

Currently, `cors_origin` is configurable so we can set that to include both, but it would be nice if the charm would automatically do this for us. This reduces the chance of human error where updating the other settings would make it less of an issue or miss updating `cors_origin`.

For `external_hostname`, I think it would need to take into account `force_https` to see if it should add `https://` or just `http://`.

For `s3_cdn_url`, it would add it as is.

srbouffard:
### Overview

This PR introduces a new optional config option, augment_cors_origin, to improve the management of CORS origins for Discourse.

## Changes Introduced

- **New config option: `augment_cors_origin` (default: `true`)**
  - When enabled, this option *augments* the user-defined `cors_origin` value with:
    - The configured `external_hostname` (prefixed with `http://` or `https://` depending on `force_https`)
    - The configured `s3_cdn_url`

- **Behavioral adjustment to `cors_origin`:**
  - **New config default: `cors_origin` (default: `""`)**
  - If `cors_origin` is explicitly set to `"*"`, augmentation is **skipped**.
  - If `cors_origin` is an empty string `""`, only the augmented values are used (if augmentation is enabled).

### Rationale

This change helps reduce configuration mismatches and human error in setups. It makes the charm's CORS behavior easier to manage, while maintaining backward compatibility.

### Checklist

- [ ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)
- [ ] The changelog is updated.

Closes #218


src/charm.py
```diff
@@ -242,6 +242,36 @@ def _get_external_hostname(self) -> str:
...
+            ext = self.config.get("external_hostname")
+            if ext:
+                scheme = "https" if self.config.get("force_https") else "http"
+                origins.add(f"{scheme}://{ext}")
+
+            cdn = self.config.get("s3_cdn_url")
+            if cdn:
+                origins.add(cdn)
+
+        return ",".join(sorted(origins)).lower() if origins else ""
```
amandahla, line 273: minor: why .lower()?

config.yaml
```diff
@@ -1,9 +1,19 @@
 # Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 options:
+  augment_cors_origin:
```
amandahla, line 4: I thought that this should be true by default (without even needing a configuration). Should it be false in what scenarios?

tests/unit/test_charm.py
```diff
@@ -844,3 +844,134 @@ def test_setup_and_activate_on_upgrade(monkeypatch: pytest.MonkeyPatch):
...
+                "force_https": True,
+                "s3_cdn_url": "https://foo.com",
+            },
+            "https://foo.com",
+            ActiveStatus(),
+            id="Duplicated origins across cors_origin and augmentation",
+        ),
+    ],
+)
+def test_get_cors_origin_behavior(config, expected, status):
```
amandahla, line 964: What do you think about adding the new unit test using ops.testing/scenario?

DeeKay3: I agree, I think this would be a great practice too.

pyproject.toml
```diff
@@ -18,6 +18,10 @@ show_missing = true
 [tool.pytest.ini_options]
 minversion = "6.0"
 log_cli_level = "INFO"
+pythonpath = [
```
DeeKay3, line 21: Is this required for pytest to run?

src/charm.py
```diff
@@ -242,6 +242,36 @@ def _get_external_hostname(self) -> str:
...
+            Comma-separated CORS origins string.
+        """
+        user_value = self.config.get("cors_origin", "").strip()
+        if user_value == "*":
+            # No need to augment if all origins allowed
+            return "*"
+
+        origins = set()
+        if user_value:
+            origins.update(o.strip() for o in user_value.split(",") if o.strip())
```
DeeKay3, line 261: Minor suggestion, feel free to reject, or use a changed version:
```suggestion
            origins.update(s for s in (o.strip() for o in user_value.split(","))if s)
```
This calls `o.strip()` once. `s` stands for stripped.

src/charm.py
```diff
@@ -263,10 +293,15 @@ def _is_config_valid(self) -> bool:
...
 
-        if missing_fields:
-            errors.append(f"Required configuration missing: {','.join(missing_fields)}")
+        if (
+            self.config.get("enable_cors")
+            and self.config.get("cors_origin") == ""
+            and not self.config.get("augment_cors_origin")
+        ):
+            errors.append(
+                "invalid CORS config. Either `augment_cors_origin` must be enabled or `cors_origin` must be none-empty"
```
DeeKay3, line 303:
```suggestion
                "invalid CORS config. Either `augment_cors_origin` must be enabled or `cors_origin` must be non-empty"
```

src/charm.py
srbouffard: The main motivation behind this was that since both the host and scheme are case insensitive, then applying a `.lower()` would normalize the output and improve consistency.
However, on further digging it appears that a case-sensitive match of the `Origin` value is part of the ([spec](https://www.w3.org/TR/2020/SPSD-cors-20200602/), even if uncommon.
So I have removed the lower() to retain the exact casing specified by the user-defined config.
Good catch! :)

tests/unit/test_charm.py
srbouffard: Could you point me towards a good example? I have not worked with ops.testing/scenario. I based those tests on others I found in the file but if there were better tests to base myself on I will gladly refactorize to them :P

DeeKay3: Some examples from other charms:

https://github.com/canonical/maubot-operator/tree/main/tests/unit
https://github.com/canonical/synapse-operator/tree/2/main/tests/unit_ops

I would suggest writing it in a new folder called `unit_ops` as in synapse-operator. Happy to help throughout any time, although I am still a newbie too on scenario :D

config.yaml
amandahla: Good points. However, considering that we have the enable_cors configuration, I suggest following the same approach used in the [workload](https://meta.discourse.org/t/setup-cross-origin-resource-sharing-cors/270819) by leaving cors_origin empty. If it is set, then you can use it. By the way, it's not recommended to leave "*" as the default.
Another suggestion would be blocking the charm if enable_cors is set, but there is no value in "cors_origin".

tests/unit/test_charm.py
srbouffard: Tests have been moved over to using Scenario, within a `unit_ops` folder as suggested!

DeeKay3 (approved):
LGTM! Thank you!

config.yaml
srbouffard: The new behavior, based of the default config values is that CORS is enabled and auto-populated based off external_hostname and the s3_cdn. The user can then:
- Define additional `cors_origin` to be appended on top of external_hostname and the s3_cdn
- Define a specific `cors_origin` and disable the augment altogether
- Explicitly define `cors_origin="*"`, which negates the augment. In the future we may look at restricting this and simply not allowing "*" altogether but for now, it allows us to match the current behavior from the default configs.

In changing the default value of `cors_origin=""` I also changed the default of `augment_cors_origin=true` so that the CORS state based off the default config values is a valid state and wont defacto block the charm.

tests/unit_ops/conftest.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2025 Canonical Ltd.
```
amandahla, line 1: Thanks for creating a unit test with ops.testing!
Please, could you organize as mentioned in the spec "ISD214 - Task Group - Unit Testing" ?
"Move the existing unit directory to unit_harness and create the new tests in the new unit directory.  This also implies adding a new entry to tox.ini."
ps.: I noticed that the example was still mentioned "unit_ops" so I fixed it now.

tests/unit_ops/test_charm.py
```diff
@@ -0,0 +1,134 @@
...
+            },
+            "http://example.com,https://bar.com,https://cdn.test,https://foo.com",
+            ActiveStatus(),
+            id="Multiple user-defined cors_origins with augmentation",
+        ),
+        pytest.param(
+            {
+                "cors_origin": " https://foo.com , https://foo.com ",
+                "augment_cors_origin": True,
+                "external_hostname": "foo.com",
```
amandahla, line 104: Would it be possible to add one without setting "external_hostname"?

amandahla (approved):
Nice, left some minor comments.

tests/unit_ops/conftest.py
srbouffard: Changes to the folder structure of the ops and harness tests as well as updates tox to reflect those changes

tests/unit_ops/test_charm.py
srbouffard: Added a new testacase with id `Augment with external_hostname not explicitely defined`

CODEOWNERS
```diff
@@ -1 +1 @@
-*       @canonical/is-charms
+*       @amandahla, @DeeKay3, @nrobinaubertin, @srbouffard
```
amandahla, line 1: No commas :)
Also, see we are here you can change it by asking reviews for docs changes to erin except the changelog
Example:
https://github.com/canonical/maubot-operator/blob/main/CODEOWNERS

PR #341, opened 2025-06-26
feat/ISD-3609: Add multi-level category setting to charm config

DeeKay3:
Applicable ticket: ISD-3609

### Overview

Added `max_category_nesting` config option, which allows the configuration for the number of nested categories one can create. This is a Discourse option, this PR only enables the ability to change the configuration option in the Discourse app through the charm.

### Rationale

This was a requested feature to be implemented so that more nested categories could be added.

### Checklist

- [X] The documentation for charmhub is updated.
- [x] The changelog is updated.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -37,7 +37,7 @@ options:
     default: false
   max_category_nesting:
     type: int
-    description: "Maximum category nesting allowed."
+    description: "Maximum category nesting allowed. Minimum is 2, maximum is 3."
     default: 2
   saml_sync_groups:
     type: string
--- a/config.yaml
+++ b/config.yaml
@@ -35,6 +35,10 @@ options:
     type: boolean
     description: "Force SAML login (full screen, no local database logins)."
     default: false
+  max_category_nesting:
+    type: int
+    description: "Maximum category nesting allowed."
+    default: 2
   saml_sync_groups:
     type: string
     description: "Comma-separated list of groups to sync from SAML provider."
--- a/src/charm.py
+++ b/src/charm.py
@@ -465,6 +465,7 @@ class DiscourseCharm(CharmBase):
             "DISCOURSE_DEVELOPER_EMAILS": self.config["developer_emails"],
             "DISCOURSE_ENABLE_CORS": str(self.config["enable_cors"]).lower(),
             "DISCOURSE_HOSTNAME": self._get_external_hostname(),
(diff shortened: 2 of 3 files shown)


config.yaml
```diff
@@ -35,6 +35,10 @@ options:
     type: boolean
     description: "Force SAML login (full screen, no local database logins)."
     default: false
+  max_category_nesting:
+    type: int
+    description: "Maximum category nesting allowed."
```
srbouffard, line 40: Should we add a warning here regarding the potential disruptive behavior of scaling down the nesting? From what I can gather, the operation is not destructive, but it would mean some categories are moved up a level or inaccessible.

srbouffard (approved):
Small comment, else LGTM!

config.yaml
DeeKay3: I tested this behaviour. There is no disruptive behavior, the already created categories stay as they are, downscaling only affects future category creation.

src/charm.py
```diff
@@ -465,6 +465,7 @@ def _create_discourse_environment_settings(self) -> typing.Dict[str, str]:
             "DISCOURSE_DEVELOPER_EMAILS": self.config["developer_emails"],
             "DISCOURSE_ENABLE_CORS": str(self.config["enable_cors"]).lower(),
             "DISCOURSE_HOSTNAME": self._get_external_hostname(),
+            "DISCOURSE_MAX_CATEGORY_NESTING": str(self.config["max_category_nesting"]),
```
amandahla, line 468: Should this value be validated? Is there a min/max for it?

DeeKay3: Discourse [mentions](https://github.com/discourse/discourse/blob/5bbdc8a813caf55ab3147ac65b5ffafb5e0aab90/config/site_settings.yml#L281) min=2 and max=3. But, technically, you can set the value however you want, you will just see funny behaviors when you set the value to:

<=0: You get an error, but the category is still created. You can't create any subcategory at all though.
1:  This works as intended, means you can create categories, but no sub-categories. The error message still says "can't create sub-category under another", but it works as intended.
>3: Weirdly, you CAN create nested subcategories that are at higher level than 3, but they are created as "private", i.e. only accessible if you know the URL. They do not appear in the UI. 

I am under the impression that anything other than 2-3 is not intended and Discourse probably doesn't care about the weird behaviors because you shouldn't set them, but you can. I think we should do the check and not allow anything other than 2 or 3. I will update accordingly.

DeeKay3: done.

